PR #7, opened 2023-08-07
Devel - OenLeu Assembly Review

jpnm:
Species name: Oenanthe leucura 
Taxa: Aves
Project: ERGA-Pilot


diegomics:
Thanks @jpnm for helping us with the testing! Your pre-curation assembly will be reviewed now 😃 

As a side development note, I think we should add in the readme that when creating the pull request add something like:
```
Species name: Oenanthe leucura 
Taxa: Aves
Project: ERGA-Pilot
```

@tbrown91 would you like to check the EAR?

tbrown91:
Hi @jpnm,

It looks like you might have run purge-dups in default mode, which introduces 13bp gaps where retained haplotypes overlap. In the future it would be best to always run the get_seqs with the -e flag to avoid this https://github.com/dfguan/purge_dups/issues/31

I am a little concerned about the size difference between the two assemblies. Do you have any evidence that this species has a very large W chromosome? This is also visible in the BUSCO scores.

There is obviously a lot of unscaffolded sequence, which is to be expected from a bird genome with 41 chromosomes and a lot of micro-chromosomes. If you or someone else is happy to curate as is, then I would be happy to approve. Otherwise I would perhaps suggest not running hifiasm in the phasing mode with hic data, but running with the --primary tag to produce a primary and alt assembly. The primary assembly should then contain both sex chromosomes and be easier to scaffold due to higher contiguity. This is of course with the downside of not being phased.

Please let me know if you have any further questions

Tom

jpnm:
Hi @tbrown91,
Thanks a lot for the feedback! 
Since August gives me a chance to tap into a bunch of resources, I'm gonna take a shot at creating a different version, like you suggested with hifiasm (not phasing) and purge dups (with the -e flag).
Once again thanks!

tbrown91:
That sounds pretty good to me. 

If you recreate the EAR and push a new commit, we can continue the discussion here

diegomics:
I second all the comments from Tom. Also, I would use 21 or 31 as k-mer length for genome profiling.
Also, I'm wondering how the proceeding was for the Hap1/Hap2 purging (discarding or what).
I think the assembly is meeting EBP criteria (the Hap1), but it can be improved without much effort if possible. In that case, you can update Hifiasm to v0.19.5
As Tom mentioned, you should not create a new pull request, just update the EAR.pdf in your repo, and it should update automatically here so we can review it ;)

jpnm:
Hi @diegomics. 
I did the purge_dups following the VGP Galaxy tutorial. At the time the steps were: purging hap1, adding the purged hap1 sequences to hap2, purge hap2. I notice a huge change in the tutorial. It looks much more simplified now, so I will check it before producing the new assembly!
Thanks to both

diegomics:
Ok, now I get a better picture! For a HiC-phased assembly, the idea is to purge Hap1 and discard the dups and retained haplotypes, purge Hap2 and discard the dups and retained haplotypes. If you have a Pri/Alt assembly, then you need to do the double purging: purge Pri, merge the dups and retained haplotypes to Alt, and then purge that merged asm. We need to work on tutorials (one thing at a time!)
Feel free to ask further questions here :)

jpnm:
However it can explain little differences, the major one is given by the hifiasm output!

diegomics:
Thanks for the update! I see that you uploaded two EARs for the same assembly. One is based on a HiC-phased strategy (hap1 & hap2), and the other is Hifiasm-solo (Pri & Alt). Is this correct?

diegomics:
I think the decision should be based on Hap1 against Pri (because Hap2 is not good). Between those two, Pri is the best one. @tbrown91 what do you see here?

jpnm:
> Yes. They were both done with hifiasm v0.19.5 and purge_dups were performed for both!

tbrown91:
@jpnm do you know the sex of the sampled individual? It looks like one of the sex chromosomes may be in the alt assembly in your PRI version. It looks similarly as if the two hap assemblies have sex chromosome sequences in. This is a question for curation really, but it might be best to either try and pull the W/Z from the alt into the pri, or the same for hap2 to hap1.

My gut feeling in this instance is to try and get the missing sex chromosome from hap2 and move it to hap1 during curation. I know the contiguity is a bit lower than the PRI assembly, but I think there are advantages to having the phased sequences, without any seeming drop in BUSCO.

The purging seems to have done a pretty good job here. I assume you added the purged sequences from the primary to the alt and then ran purging again on this set of "alt" sequences. Otherwise I can't think of a way for the kmer completeness to go up for the alt assembly. Hopefully this will produce a good reference and a set of phased alternate autosomes.

I suspect, based on the pretext map, that this will not reach 90% scaffolded into chromosomes, but for this species with micro-chromosomes, I think this will be the best possible.

jpnm:
> @jpnm do you know the sex of the sampled individual? It looks like one of the sex chromosomes may be in the alt assembly in your PRI version. It looks similarly as if the two hap assemblies have sex chromosome sequences in. This is a question for curation really, but it might be best to either try and pull the W/Z from the alt into the pri, or the same for hap2 to hap1.
> Its a female, so the heterogametic sex. 

> My gut feeling in this instance is to try and get the missing sex chromosome from hap2 and move it to hap1 during curation. I know the contiguity is a bit lower than the PRI assembly, but I think there are advantages to having the phased sequences, without any seeming drop in BUSCO.
> I understand. I will try to check it!

> The purging seems to have done a pretty good job here. I assume you added the purged sequences from the primary to the alt and then ran purging again on this set of "alt" sequences. Otherwise I can't think of a way for the kmer completeness to go up for the alt assembly. Hopefully this will produce a good reference and a set of phased alternate autosomes.
> Yes, was exactly what I did.

> I suspect, based on the pretext map, that this will not reach 90% scaffolded into chromosomes, but for this species with micro-chromosomes, I think this will be the best possible.

tbrown91:
@diegomics for me this is now a pass. What do you think?

diegomics:
Yes. I agree with Tom's review that the HiC-phased version is better considering the next step (manual curation). Since this is a pre-curation review, we are good to go. Thanks @jpnm!

diegomics (changes requested):
@jpnm please remove the bOenLec1_PRI_EAR.pdf so we can close this PR ;)

tbrown91 (commented):
All good.

Aim to add missing sex chromosome from hap2 into hap1 during curation and treat this as the "primary" assembly

PR #8, opened 2023-09-08
Devel - CorAlb Assembly Review

jpnm:
Species name: Corema album
Taxa: Embryophyta
Project: ERGA-Pilot


tbrown91:
Hi Joao,

I'm a little concerned about the HiC library here. Are you able to get hold of the following statistics?

Total number of read pairs/total coverage
Number of mapped reads
Number of cis/trans interactions
Number of cis interactions >1kb
Number of cis interactions >10kb
Number of cis interactions >50kb

This also looks like a pretty interesting case of sex chromosomes in flowering plants. I guess this is a male plant as it looks like there are some heterogametic chromosomes in the merqury plots. Does this sound right to you?

@nadegeguiglielmoni I feel a little uneasy about the purging here, what do you think?

nadegeguiglielmoni:
@tbrown91 Because of the k-mer spectra? I would try changing --xmax and --ymax.

jpnm:
Hi Tom and Nadege,

I wanted to drop you a quick message to say sorry for taking so long to get back to you. I've been swamped with BGE tasks that needed my urgent attention. Because of these commitments, I'll need a few more weeks to find the right time to tackle the issue you've mentioned properly.

Thanks a lot for your feedback. I want you to know that as soon as I have some free time, I'll put all my energy into sorting this out in the best way possible.

I'll be in touch as soon as I have any updates.
Cheers

tbrown91:
Hi @jpnm, not a problem, take your time. It's better everything is done correctly, rather than quickly.

PR #9, opened 2023-09-11
Older BUSCO uses "number of species".

gkanogiannis:
I could not get `make_EAR.py` to run and realized it brakes because my input BUSCO short summaries use the wording "The lineage dataset is: blabla. Creation date: date1, number of **species**: xy, number of BUSCOs: xz" instead of **genomes** . I used quast, that uses an older BUSCO version (3.0.2).

### DIFF
--- a/make_EAR.py
+++ b/make_EAR.py
@@ -413,7 +413,11 @@ def make_report(yaml_file):
                 if lineage_info:
                     return lineage_info.groups()
                 else:
-                    return None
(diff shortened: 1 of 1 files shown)


diegomics:
Hi @gkanogiannis thanks for the PR, I will review it. Please update the version of BUSCO, 3.0.2 is quite old. What database are you using?

gkanogiannis:
I am using the BUSCO from within quast (5.2.0) in a conda env. I cannot upgrade to any BUSCO >3.0.2 (breaks dependencies), so I just hacked the lineage database to use an odb10.

diegomics (approved):

PR #13, opened 2023-09-22
Devel - LepGra Assembly Review

jpnm:
Species name: Lepus granatensis
Taxa: Euarchontoglires
Project: ERGA-Pilot


diegomics:
Hi @jpnm, thanks for the EAR :)

Do you know the sex of the individual?
Just a heads-up, the yaml and make_EAR.py script were updated to include this information now. If you want to have this in the pdf, please check the updated yaml (you can re-use the one you have filled, only thing to add is the `Sex:` line) and download and run the updated make_EAR.py. The conda environment is the same ;)

tbrown91:
Hi @jpnm, this one looks really good. It looks like the X chromsome is in hap1 (scaffold 7) and from the kmer profiles it looks like there is some Y-chromosome sequence in hap2, but I suspect this is not visible in the pretext map.

The BUSCO is quite low for all other lagomorpha assemblies that I could find, so I'm not too concerned about the scores. You could perhaps "artificially" inflate this by using another database, such as vertebrata or mammalia.

Please confirm that it is indeed male, and then I'm happy to pass this one straight away.

jpnm:
Hi @diegomics and @tbrown91,

Thanks for your comments.
The individual is indeed a male! I already replaced the report with the new version containing the sex information.
Concerning the BUSCO scores, I've produced a report with the mammalia db but although it improves for the hap2, it's worst for the hap1. So I think I will keep it as it is now.

tbrown91 (approved):

PR #16, opened 2023-10-19
EAR for mApoAgr2 pre-curation

talioto:
Species: Apodemus agrarius
Taxa: Rodentia
Project: ERGA-BGE


tbrown91:
Looks good to me

It's a shame again that it looks like only one sex chromosome is in the assembly. If I were intelligent enough I'd think of better ways of handling sex chromosomes in ONT reads.

tbrown91 (approved):

diegomics (approved):
All the EBP metrics are met, and the assembly looks good to go for manual curation. Thanks and congrats to the CNAG team ;)

talioto:
WRT sex chroms, they are both there. Thus my comment to Diego about resolution of the map files. I would ask for the pretextmaps, too, apart from the pdf report. screenshot attached here.
Screenshot 2023-10-19 at 15 53 39

PR #54, opened 2024-08-05
Add EAR bot

arash77:
This will add the bot for the EARs reviewing process.


diegomics (approved):
Massive praise to @arash77  and the Freiburg team for working on this! We were testing it for some time now, but let's see how it unfolds in the real world :)

PR #64, opened 2024-09-05
Use reviewer institution instead of researcher

arash77:
This is an update for https://github.com/ERGA-consortium/EARs/pull/54


diegomics:
This should solve the issue of the bot writing the incorrect reviewer institution in the _EAR_reviews.csv_ after merging via applying `reviewer_data.get("Institution", "")`
Thanks for the quick fix @arash77 🙌

PR #65, opened 2024-09-06
add daXanOrie1

andar27:
## Assembly review request

- ToLID: daXanOrie1
- Species: Xanthium orientale
- Project: ERGA-BGE
- Affiliation: Genoscope


tbrown91:
ok

talioto:
Yes

talioto (approved):
Looks great. I don't think any changes are necessary, so I Accept. BUSCO dups don't seem to correlate with any hapdup signals. BUT, I'd like to understand better some things:
1. How should I interpret the contigs extension track?
2. There is no gap info - I suppose I need to learn the answer to 1.
3. Omni-C coverage is a bit low, but still enough for scaffolding and confirming overall chr structure. Is this normal for Genoscope or was there an issue with the library?

tbrown91:
Thanks @talioto 

@andar27 can you let us know what the extensions track refers to and about the HiC?

It does look a little sparse, but this assembly was so contiguous, that I guess it was not a huge issue.

Great job and I will merge once you reply to Tyler's queries

andar27:
Hi @talioto and @tbrown91, thanks a lot for your review.

The contigs extension track give the same information than the gap track. We create it for more visual information about the contig delimitation.
For the HiC coverage, only one library was sequenced (the yield was lower than expected, 79M reads instead of >100M). We tested before generating a second library, and as this genome posed no particular problems, we decided to stop sequencing.

Adama

PR #66, opened 2024-09-09
add laCymNodo2

bistace:
## Assembly review request

- ToLID: laCymNodo2
- Species: Cymodocea nodosa
- Project: ERGA-BGE
- Affiliation: Genoscope


diegomics:
Ok

SarahPelan:
Yes

SarahPelan:
Hi @bistace,
I have looked through laCymNodo2 and here's what I think:

SUPER_2
746-768 Kbp possible small retained haplotype
I think your SUPER_2_unloc_1 can join in the gap @ 18.3Mb in -ve orientation

SUPER_3
26.63 - end belongs on RH end of SUPER_5 in other orientation

SUPER_7
9.408-9.424Mb, possible small retained haplotype
Break in gap @ 19.97Mb and join this piece to LH end of SUPER_7

SUPER_10
4.55-4.59Mb approx, looks like retained haplotype

SUPER_12
10.891-10.907Mb, possible small retained haplotype

SUPER_13
5.35-5.37Mb, possible small retained haplotype

SUPER_15
Looks like incorrect join @ 3.28Mb, flip from 0-3.28Mb and join to RH end
scaffold_31 can be placed on SUPER_15, and potentially goes in the gap created from the above break. This isn’t 100% certain though as it could go in at either gap at 2.69, or 2.63 so might be better left as unlocalised.

SUPER_16
Break @ 6.22Mb, flip 0-6.22 and rejoin

SUPER_17
0-236Kb approx looks like it should move to SUPER_17 @ approx 2.79Mb
3.976Mb - I think this looks like the telomeres are joined so needs breaking here, and 3.97 to end moving to start (there is affinity at this point to other telos)
10-4-10.6 approx looks like it needs to be flipped

SUPER_18
I think this needs a bit more investigation:
If the telomeric sequence at the start of SUPER_18 is definitely correct, then the hic mappings indicate that there is also potentially telomeric sequence @ 1.428Mb (hic mapping indicates they may join)
I would suggest to break @ 1.428Mb, and then flip from 1.428 to end and rejoin. The resulting new join looks to have a similar amount of hic data mapping to it in the map to the current one @ 1.428Mb.

Thanks,
Sarah

diegomics:
Hi @bistace,
Please take a careful look at the excellent review from @SarahPelan. Related to SUPER_18, hopefully regenerating the plot after the flipping will fix the Telo signal.
A few things I wanted to point out:
. Why the QV and Kmer completeness in the pre-curation assembly are showing such a low values?
. It is nice to have the unloc at the end on begginign of each chromosome

Just a heads-up. We updated the make_EAR and yaml files, now is easier to enter the Assembly and Curation tools used :)

bistace:
Hello and thank you for your review!

I agree with all your suggestions and have made the changes. The Pretext map and the EAR are being generated and should be ready tomorrow.

@diegomics It seems that I gave the wrong input directory for the pre-curation stats, this will be fixed in the next iteration of the EAR report. I agree with the unlocs needing to be at the end of chromosomes, the 'preserve order' checkbox was apparently not enabled when I generated the AGP and this will also be fixed with the corrected map. Thank you for updating the make_EAR script and yaml, I will need to integrate it into our pipeline so this will be a bit slower than the rest!

bistace:
Hello and sorry for the delay! 

I did the modifications on the SUPER_18 scaffold but the telomeric signal did not appear and the contacts seem to indicate that the region is flipped. The updated Pretext map includes this modification but I think that it would look better by flipping back the region.

diegomics:
Thanks for generating the EAR PDF with the new script @bistace !!
I think the super 15 looks better like this:

Screenshot 2024-09-18 at 14 45 14


Let's wait for @SarahPelan feedback :)

SarahPelan:
Hi @bistace 
Apologies for the delay in re-looking at this one - I was off last week.

I think:
SUPER_3
Still think the rh end looks incorrect  and 26.63 - end of SUPER_3 belongs on RH end of SUPER_5 in other orientation

SUPER_10
4.55-4.59Mb approx, looks like retained haplotype still - if you’d prefer to leave as is then that’s fine

SUPER_15
10.26-13.53Mb (end of chr) needs flipping as @diegomics  says

SUPER_16
Still need to break @ ~6.2Mb, flip 0-6.22Mb and rejoin

SUPER_18
I agree that the region at the start looks like it should be flipped back.

Also, I noticed there is scaffold_31_unloc_1 - It belongs in SUPER_15 - possibly a scaffold painting issue in pretext - I frequently do this!!

Thanks,
Sarah

bistace:
Hi @SarahPelan, thank you again for your review and no worries for the delay, I hope that you had a great time!

@diegomics nice catch for the SUPER_15, I did not see it!

I've updated the EAR and Pretext map to include your modifications. For the SUPER_3, I've moved only a portion of the region to the end of SUPER_5 because I think that, even though it shows some contacts, moving the entire region does not make a nice diagonal appear like at the end of SUPER_3.

diegomics:
Hi @bistace please check the link, I think the pretext file is not updated

bistace:
Hi, I reuploaded the map so the link should work correctly now!

SarahPelan:
Hi @bistace 
I have rechecked, I think you missed off some unloc tags?
SUPER_2
27.4-end of chr, either needs to be broken and tagged as unloc, or joined in @ ~18.3Mb

SUPER_5
25.5-chr end, needs to be in other orientation.  Then, the addition of the remaining  small piece from the end of SUPER_3  looks good across the diagonal
laCymNodo2_SUPER_5_SUPER_3_join


SUPER_7
24.1-end of chr,needs to be broken off and tagged as unloc

SUPER_10
18.9-end of chr,needs to be broken off and tagged as unloc

SUPER_17
Looks like there is some retained hap between 6.4-6.6Mb

Also, SUPER_18
This now has a telo tag @ ~1.4Mb, sequence from ~1.4Mb onwards appears to have contacts to other telomeres so I think it is probably telomere here.
I think that there is probably telomere on both sides of the gap, and the piece from 0-1.4Mb actually belongs on the other end (same orientation).  Would you be able to check?  This doesn’t look great on the hic map when it’s moved, but there could be data missing from this new resulting join between ~10.37Mb and 0Mb. 
If moved, I think the overall patterning of the contacts across SUPER_18 then looks more consistent with the other chromosomes, I've added a screenshot and circled the patterning I mean - I hope it makes sense

SUPER_18_current
SUPER_18_proposed
hic_contact_patterns


SUPER_19 painted by mistake as a chromosome?

Thanks,
Sarah

bistace:
Hi @SarahPelan,

I checked the SUPER_18 situation and I agree with you that moving the start to the end makes the contacts look a bit more like other chromosomes. 

I searched for telomeric repeats and they appear to be located before the gap (I found only three TTTAGGG repetitions 6kb after the gap). Do you think that the region before the gap should still be moved at the end of the chromosome?

SarahPelan:
Hi @bistace 
Yes, I think that whole section 0-1.4ish of SUPER_18 should move to the end.
Even though there are telo repeats only before the gap, the hic contacts suggest that the sequence after the gap is also telo.  We quite often see these mis-joins around the telomere with it looking as though the join should be where the assembler has placed it, but due to the telo tags, we can be confident that these in fact are the ends.
I hope that make sense..
I think it's likely that there is data missing across the new join which is why you aren't getting the nice diagonal mappings.
Thanks,
Sarah

bistace:
Ok, I will move this section then and I will regenerate the EAR and Pretext map! Thank you again for your review and very helpful comments.

SarahPelan:
no problem :)

bistace:
Hello,

I applied the requested changes and updated the EAR report and Pretext map. After moving the part at the end of SUPER_3 to SUPER_5, the end of SUPER_3 now shows contact with the end of several other chromosomes so I moved the part you indicated @SarahPelan and did not apply any further modifications to  the SUPER_3 scaffold.

SarahPelan (approved):
Hi @bistace 
This now looks great, thanks for making all those changes!
Sarah

diegomics:
Thanks @bistace for another nice Genoscope assembly! I want to take a moment to point out the extraordinary review by @SarahPelan 🙌

PR #72, opened 2024-09-18
add Diadema_setosum

Arno75:
## Assembly review request

- ToLID: eeDiaSeto1
- Species: Diadema setosum
- Project: ERGA-BGE
- Affiliation: Genoscope


tbrown91:
ok

gitcruz:
Yes

gitcruz:
Hi @Arno75 

I went through the EAR and the contact map, congratulations I think you've done an excelent work. Specially, given the low-density of contacts for your map  (could you tell me what was the total number of valid pairs ?) and background noise observed ( I guess this is not a hires map right)). You introduced 727 gaps during curation and your drastically reduced the %D buscos!!!! The chromosomal territories are clearly defined for each SUPER (still called scaffolds here, because you haven't painted them yet). I agree that the shortest scaffolds (23-98) should remain unplaced. 

@tbrown91 I'd say this genome is ready for ENA upload without any further review. Not sure is totally necessary to paint it.

For my own interest, I'd like to know:
1. Which tool you've used to produce the TAPAs Graph Summary shown in the Post-curation contamination screening.
2. What is the difference between find_telo and telomere tracks. In this case, it seems that find_telo always place them at terminal positions while telomeres are "subtelomeric" in some cases.

Thanks,
Fernando

tbrown91:
Thanks @fcruz, sounds good to me!

@Arno75 congrats on a great assembly. Could you please answer some of Fernando's questions and then we will merge the genome into main

Arno75:
Hi @fcruz,
Thanks for the review. I'm glad to see the scaffolding curation is good (I can do the scaff painting if necessary).

1- Concerning the "TAPAs" graph, it's a graph generated in R resulting 
from the combination between %GC, coverage (assembly reads) and a taxonomic assignment made from 
genes detection (with metagen).
2- I'm not sure I understood the question correctly. 
But the FindTelo tool allows us to determine the positions of the telomeres which we report in 
the "telo track" available under pretext. This detection is not always reliable, especially when 
the patterns are complicated to determine.

Arnaud

gitcruz:
Hi Arnaud,

 @Arno75 see my replies below in _italic_

> Hi @fcruz, Thanks for the review. I'm glad to see the scaffolding curation is good (I can do the scaff painting if necessary).
_For me is not totally necessary, but we routinely paint the larger scaffolds that we think correspond top chromosomes so they show as SUPER in the final assembly distinguishing them from the unloc (in your case are scaffold_23 to 98) and the alternative haplotigs (I guess your remove them because I don't see them). I think is nice to see the SUPER and shows it's chromosome level. What do you think  @tbrown91?_
 
> 1- Concerning the "TAPAs" graph, it's a graph generated in R resulting from the combination between %GC, coverage (assembly reads) and a taxonomic assignment made from genes detection (with metagen).

I see, it looks similar to what BTK does. could you point me to the exact metagen github or doi publication? I searched for it and found similar names sucha as CCmetagen, but not simply metagen.

 2- I'm not sure I understood the question correctly. But the FindTelo tool allows us to determine the positions of the telomeres which we report in the "telo track" available under pretext. This detection is not always reliable, especially when the patterns are complicated to determine.
_In your pretext map there are two different extensions one called "findtelo" reporting the positions as you mention above and another one called "Telomeres". For scaffold_2, they report telomeres at different positions at the end. What is the difference between these two tracks? again what is thelink to this tool is it the script called findtelomeres here: https://github.com/JanaSperschneider/FindTelomeres?  _
> 
> Arnaud
Thanks,
Fernando

bistace:
Hi Fernando,

TAPAs is a tool that we use internally to assign a taxonomic rank to each contig/scaffold. We first do a gene prediction with metagen (https://metagene.nig.ac.jp/) because it is fast and we do not need to be very precise with this step. When we have all gene predictions, we assign a taxonomic rank to them by doing a mapping against NR. Then, we assign a taxonomic rank to each contig/scaffold by using its gene predictions taxonomic assignments and we apply a weighted-LCA to find the taxonomic rank that will be given to the contig/scaffold. 

The `findtelo` track is also an internal tool that we use and that looks for common telomeric repeats at the edge of sequences.

Please tell me if you have some more questions and I'll let Arnaud answer you for the number of valid pairs.

Benjamin

gitcruz:
Thanks Benjamin,

What is the difference between the findtelo and telomeres extensions in
Your contacta map?

I would like to know the total valid pairs or equivalent HiC Contacts (It
vary between preprocessing toos but should be basically the total pairs
used by Your scaffolder)

By the way, which hic scaffolding tool have you used ...i think i missed
that detail.

Best regards,
Fernando

Arno75:
Hello fernando,
Find telo only looks at telomeric repeats at the beginning and end of the scaff while telomere looks for repeats everywhere on the genome. I'm looking for valid pairs and i will tell you.

Arnaud

gitcruz:
Hi @arno75

Ok that explains it. Then *findtelo* simply confirms the presence at
chromosome ends while *telomere *can give information about chromosomal
mixups.

Yes, please. I'm just curious to know. It looks that you had few contacts
to work with.

Thanks
Fernando

tbrown91:
@gitcruz could you please "Approve" your review, then I can merge and close this review

Many thanks all

gitcruz (approved):

gitcruz:
Sure It's approved now. Congratulations again @Arno75

gitcruz:
@Arno75 @bistace 

Hi Arnaud and Benjamin,

My suggestion: For the shake of transparency and reproduciblity, I think you should consider sharing these tools in github or some other public server. 

Best,
Fernando
> Hi Fernando,
> 
> TAPAs is a tool that we use internally to assign a taxonomic rank to each contig/scaffold. We first do a gene prediction with metagen (https://metagene.nig.ac.jp/) because it is fast and we do not need to be very precise with this step. When we have all gene predictions, we assign a taxonomic rank to them by doing a mapping against NR. Then, we assign a taxonomic rank to each contig/scaffold by using its gene predictions taxonomic assignments and we apply a weighted-LCA to find the taxonomic rank that will be given to the contig/scaffold.
> 
> The `findtelo` track is also an internal tool that we use and that looks for common telomeric repeats at the edge of sequences.
> 
> Please tell me if you have some more questions and I'll let Arnaud answer you for the number of valid pairs.
> 
> Benjamin

PR #73, opened 2024-09-23
add Acomys minous

SMangenot:
## Assembly review request

- ToLID: mAcoMin1
- Species: Acomys minous
- Project: ERGA-BGE
- Affiliation: Genoscope


tbrown91:
ok

tbrown91:
@SMangenot Thank you for submitting the review. In the future could you please create a separate branch in your local fork, so that you are not submitting from main? This is also stated in the EAR wiki [here](https://github.com/ERGA-consortium/EARs/wiki/Getting-your-assembly-reviewed-using-the-EAR#4)

Thanks for agreeing to review @talioto there look to be a few things to fix in the scaffolding I have already seen from a quick glance

talioto:
In summary, I'd only strongly suggest the small rearrangement on SUPER_7. All other comments express my own uncertainty. Maye @tbrown91 has additional comments???

tbrown91:
My biggest concerns are SUPER_15 and 16. 

SUPER_15: It looks to me like the region 97.9Mb-100.5Mb should be placed at 55.6Mb and the region 100.5Mb-end should be at 91.2Mb
SUPER_16: 73.9Mb-83.5Mb looks like it should be at 33.5Mb

Maybe this is a very heterozygous species and the differences between the two haplotypes are creating issues in the Hi-C map. 

SUPER_7 contig I agree
SUPER_9 looks ok to me as is
SUPER_11 there is a drop in coverage - I agree this is probably a dup that should be removed
SUPER_14 - sure, if you want to have unloc sequences
SUPER_18 & 20 look ok to me - I would not join as the main interaction comes from this repetitive block that is also in most of the unscaffolded sequences

Thanks for looking through the map @talioto 
@SMangenot please have a look through the comments and see what you think. Hopefully this will not require a lot of changes. Do you also have a more informative contamination plot? It's not very helpful to see that all sequences are "eukaryote"

talioto:
@tbrown91 , I'm not convinced with making the edits on 15. You could invert Mb 97.8 to the end and it looks better. but the two contigs in question should stay together. I think there's a bit of repetitive sequence causing some of the signal you see as evidence for translocating but I see this type of signal in other parts of the assembly, too.

As for SUPER_16, I totally agree with you. Nice catch.

tbrown91:
thanks again @talioto

SMangenot:
Thanks for your feedback. I have just submitted the new EAR file following your comments. I've made all the changes you asked for, except for SUPER_18 and SUPER_20.

tbrown91:
Thanks @SMangenot that was pretty rapid.

@talioto can you give the map one last look through. Hopefully this will not take too much effort to finalise

talioto (approved):
I think SUPER13_unloc1 should be unplaced. I don't see evidence to add it to SUPER_13. This is so minor that I think I can trust you to rename the sequence back to the original scaffold name and remove it from the unlocalized list. If @tbrown91 would like you to redo the contact map and EAR generation, that's fine, but not required by me.

tbrown91:
Thanks @talioto 

Great job @SMangenot please go forward with the upload to ENA.

PR #77, opened 2024-09-27
EAR upload for wrPlaCost1.

tommathers:
## Assembly review request

- ToLID: wrPlaCost1
- Species: Placobdella costata
- Project: ERGA-BGE
- Affiliation: Sanger


diegomics:
Ok

bistace:
Yes

bistace:
Hi @tommathers,

nice curation! The genome structure looks great overall and a lot of chromosomes contain both telomeric regions. 

However, I’ve noticed something intriguing: there are multiple small sequences (around 50kb) flanked by gaps where sequencing coverage drops by half. This pattern appears across all chromosomes and repeats dozens of times. For instance:
- SUPER_1 at 2.34Mb, 2.92Mb, 4.65Mb
- SUPER_3 at 35.33Mb

Could this be indicative of retained haplotigs? The BUSCO scores and k-mer plots don’t suggest significant heterozygosity in the genome, so I’m curious if you have any thoughts on what might explain this.

bistace:
Forgot to ask in my previous comment, is the Pretext map included in the EAR the high resolution one? If not, would it be possible for you to generate it so we could take a closer look at these mysterious regions?

mcshane:
@bistace High-res map is here: https://tolqc.cog.sanger.ac.uk/erga-bge/annelids/Placobdella_costata/assembly/curated/wrPlaCost1.1/ear/wrPlaCost1.1_hr.pretext

diegomics:
Hi @tommathers and Shane for providing the high-res, this looks like was a complicated one.
I noticed that the EAR is showing the warning about BUSCO versions or lineage datasets not being the same across results (pre and post-curation). Can you comment on this? Thx

mcshane:
@diegomics It's because we are reusing the results from the assembly pipeline for the pre-curation results. The BUSCO version is older in the assembly pipeline than in the EAR pipeline that is run on the post-curation result. Lineage used is the same. Just us trying to not redo analysis, but can rethink if a problem.

bistace:
Thank you for uploading the high-resolution pretext map! I have reviewed it as well and the genome structure looks correct on this one too. My only concern is the high number of half-coverage sequences that are flanked by gaps that I mentioned in my previous comment and that could be retained haplotigs. Could you please check some of them, as an example by aligning it to the rest of the genome, to be sure that they can stay as is?

diegomics:
Ping @tommathers, trying to close this one.
Please comment on @bistace observations.
From my side:
As flagged, this annelid assembly contains a substantial amount of gaps, also I can see many interventions. My respects to the curator!
Many chromosomes show good telomeric signals. SUPER_4 shows off-terminal telomeric signal but I couldn't find a way of improving it, so all good.
One last small thing, can you comment on SUPER_9 @7.2? It does not look like there is repeat density like you see on SUPER_2 @8.4 rep or SUPER_3 @18.9. Since you already have seen it with higlass, you may have a better idea.
Thanks!!

tommathers:
Hi. Thanks for taking look at the assembly. @bistace re. the small regions with coverage drops. I had a look at these in higlass and it looks more like sequence drop out than haplotigs (coverage actually drops more than half). I have attached a couple of images for two of the regions flagged. As they are very tiny and there is no obvious remedy I suggest leaving as is. 

image

image

@diegomics re. the bit on SUPER 9. Not sure exactly what you mean but the small bit of sequence is contiged in without a gap and as the HIC is from a different sample I don't think is appropriate to move it. If you look across the HIC map there are ambiguous hits across the map to multiple locations so probably some sort of repeat that is variable between the HiC and pacbio. Again I suggest leaving as is and pushing this one out.

diegomics:
Hi @bistace if you are happy with the answers, please approve :)

bistace (approved):
Thank you for looking at these regions and congratulations for this curation as it seems to have been a difficult one!

PR #83, opened 2024-10-02
Add Ebenus cretica

andar27:
## Assembly review request

- ToLID: drEbeCret
- Species: Ebenus cretica
- Project: ERGA-BGE
- Affiliation: Genoscope


gitcruz:
Yes

tbrown91:
@gitcruz Have you had a chance to look through the assembly? We need to get the assemblies submitted by the end of the month

gitcruz:
Hi Tom,

Sorry, I've been quite busy and I did not look into much detail yet. But
I'll do asap this week.

Best.
Fernando

tbrown91:
@andar27 I had a quick look-through and only found a couple of issues:

Your SUPER_6_unloc_1 looks to belong to SUPER_3, are you sure you labeled this correctly?
There are a couple of instances where it looks like the coverage drops to half and the Hi-C mapping goes very strange. Do you know what is contained in these regions? Did you use mapq=0 for filtering here?
image

@gitcruz I don't think this one will take you long to look through

andar27:
Thanks you @tbrown91 for your review ! I agree with your suggestion about SUPER_6_unloc_1 and have made the change.

andar27:
Thanks you @tbrown91 for you review ! 
I agree with you suggestion about the SUPER_6_unloc_1 and have made the change. I will generate the pretext map and the EAR after the review of @gitcruz.  
There are a couple of instances where it looks like the coverage drops to half and the Hi-C mapping goes very strange. Do you know what is contained in these regions?

andar27:
Thanks @tbrown91 for your review. I agree with your suggestion about SUPER_6_unloc_1 and have made the change. I will generate the Pretext map and the EAR after the review of @gitcruz ! 

I've looked at these strange areas in detail but there's nothing in particular.

No, in this case mapq=10 was used for filtering. It was difficult to do the curation with the Pretext map with mapq=0 because there weren't enough contacts.

Adama

gitcruz:
Hi @andar27 

I think in general, @andar27 has done a good job given the low density of HiC contacts. However, I have some comments about the last decission regarding  SUPER_6_unloc_1 (now SUPER_3), the contact map extensions and the contaminant detection method.

CURATION
Although, SUPER_6_unloc_1 has been suggested to be part of SUPER_3 by @tbrown91 It's strongly contacting with SUPER_4 at position 126.4 Mb together with the mitochondrion. What is the reason? Surprisingly there is not a massive peak of repeats there compared to the assembly baseline:

strong contact to SUPER_4
Screenshot 2024-10-15 at 15 07 25

This region is multimapping 

Screenshot 2024-10-15 at 15 07 51

But has no a massive amount of repeats:

Screenshot 2024-10-15 at 15 09 12

My suggestion is to leave this scaffold  as Unplaced

----
NUMBER OF GAPS:
While the report indicates that the Total gap bp after curation is 1,800bp, the contact map shows just 1 gap at SUPER_2. What is the gap size? I would expect having 18 gaps of size 100bp. Are all the SUPER actually contigs except SUPER_2?

----

CONTAMINANT DETECTION
The report says that there is no contamination however there is one match to a bacteria with GC > 60%. Perhaps, this corresponding to the mitochondrion or chloroplast?
In the other hand, the method used seems to be classifying things grossly into distinct domains (Eukaryota or Bacteria). How can you be sure that there is not any contaminant DNA from othere eukaryotes (eg. insects or fungi) in this genome? In facta there is a cloud of blobs at coverage >1000x with GC 45-48% while the main assembly blob shows 100x and GC 33-36%. How confident are you in the TAPAs classification to rule out contamination?
 

Thanks,
Fernando

diegomics:
Thanks all for your comments. I think we can close this pretty soon.
@andar27 please, explain the contamination pointed out in the blob/tapas plot (remember that you have fields in the yaml to write down comments and other curator notes to help the reviewers evaluate the assembly faster ;))
Related to the gaps, if they are very small, they will not show, unless you use a factor to multiply the gap length in the bedgraph. Is that the case, many tiny gaps?
I agree with Tom, I will rename the SUPER_6_unloc_1 to SUPER_3_unloc_1 (and place it at the end of SUPER_3)

diegomics:
@gitcruz option of renaming SUPER_6_unloc_1 as scaffold_xxxxx is also fine in my opinion, is more conservative, and totally fine here. Please let us know what you think :)

gitcruz:
Hi @andar27,  

If the assembly does not contain contaminants and the only change was unloc  SUPER_6_unloc_1. Then don't waste your time in generating a new report. Please go ahead and generate a new fasta with this piece as unloc for ENA upload. Do you @tbrown91 and @diegomics agree?

Thanks for the clarifications about TAPAs. My only comment then is that as you are using proteins (nr) to classify gene predictions, then the taxonomic ranking is based in highly conserved stuff...How often do you get some false positives? That was the case for me with BTK when hits were relying solely in diamond hits to uniprot...

Congratulations for your new chromosome level assembly!
Fernando

diegomics:
@andar27 confirm that the contamination shown in tapas is not present in the assembly. Please notice that you have a "Contamination notes:" section where you can write down this. It is important to write some comments if something missleading is showing on a plot. So, if that's the only thing to change to the report pdf, should take 5 minutes to regenerate it.

andar27:
Hi !
False positives are very rare in our analyses. Nevertheless, for small sequences with few genes, taxonomic classification may not be very precise.  Databases may also contain contamination, which could explain some false positives.

I've already run the pipeline to generate the EAR report with the unloc modification and a new contamination plot without the bacteria. I'll update the EAR report on the github when the pipeline is finished.

Thanks for the gap extension suggestion, we will work on it to fix the issue.

Adama

diegomics:
Hi @andar27,
Sorry if I'm not getting you correctly, but if you only renamed that small scaffold, you don't need to regenerate the pretext map.
Also, if the bacteria showing in the plot was not present in the assembly, you don't need to create the plot again to fix it, just explain in the contamination notes that the plot is showing a bacteria that was removed.
Again, sorry if I'm not getting you here :)

diegomics:
Thanks for the update @andar27 
If you are happy with the EAR, please approve @gitcruz

gitcruz (approved):
Hi @andar27 

It's approved. But make sure that you use the ToLID **drEbeCret1** and not just the prefix before you upload the assembly to ENA.

Congratutations! 
Fernando

PR #84, opened 2024-10-03
Add Pipistrellus hanaki

SMangenot:
## Assembly review request

- ToLID: mPipHan1
- Species: Pipistrellus hanaki
- Project: ERGA-BGE
- Affiliation: Genoscope


tbrown91:
ok

tbrown91:
@SMangenot Could you please add BUSCO scores from a more appropriate lineage, for example mammalia or laurasiatheria? I am concerned about the number of duplicated genes and that they didn't seem to decrease even though you removed quite a number of sequences during the curation

talioto:
Yes

talioto (changes requested):
I was not quite done, but here are my notes so far:

It is a male specimen: https://www.ebi.ac.uk/biosamples/samples/SAMEA115120470 so I assume XY, although it could be XXY or X0. Other species in genus have 104 Mb X and 4 Mb Y. or 106 and 6 Mb. 

Contact density is in general very low, making decisions somewhat difficult. It's too bad the library was not sequenced to higher depth.

Telomeric and subtelomeric sequence is often incorporated but there is a lot left in the chaff/shrapnel at the end with non-specific signal. I don't know how much to trust the placement of this stuff. This is why my review has taken some time. 
SUPER_1: some faint telomeric sequence in the middle around 91.1-91.8 Mb. BUT, it seems to be contiged and contacts support this being some relic signal from a chromosome fusion. Perhaps nothing to do here.

join SUPER_7 and SUPER_5?  Similar to signal seen in SUPER_1. There is a telomeric repeat region, though.

SUPER_8: 6.5-7 Mb contig repeat that is maybe misplaced. Better to unloc it.

SUPER_14: 0-3.8 Mb. I don't think the signal is specific enough to keep this attached to this chromosome. Based on pattern of contacts to other subtelomeric regions, it seems like this is the odd one out. II would place it in the chaff.

SUPER_15: beginning subtelomeric region. I'm not sure I trust YaHS in putting this together. 

SUPER_17: interior telomeric region around 32-26 Mb. Not sure how to handle.

"Y" should probably be X. 
"Y": interior telomeric region 54.3-61.8 Mb. I don't think there's enough support to keep it in the middle of the X. This is the actual Y sequence. Keep the part next to it that has contacts with the X and is higher coverage. This is likely part of the PAR. See my savestate.

Perhaps minimap2 alignments to other species in the genus would help sort some of this out.

talioto:
Here's my savestate: https://ccnag-my.sharepoint.com/:u:/g/personal/tyler_alioto_cnag_eu/EfjGxTn7WFhGlJ8ovoUgzvsBCqk3wB01soLHeweT8NyVUQ?e=4fDcGT

tbrown91:
Thank you for the review @talioto!

@SMangenot can you please look through Tyler's suggested changes and see if they make sense in the context of the Hi-C map. I wonder if looking at synteny to other pipistrellus genomes would also help, e.g. these 4 are all in chromosomes, including the kuhlii which is stated as scaffold: https://www.ncbi.nlm.nih.gov/datasets/genome/?taxon=27671&reference_only=true

talioto:
It would be good to know if there's been any progress on this. We're in a time crunch here. Could use this genome span.

ldemirdj:
Hi @talioto, 

I'll reach out to Sophie for an update, but she's still working on the map. We should have news soon, and we'll be able to submit it before the October 31st deadline.

Thanks! 

Lola

ldemirdj:
Hi everyone, 

Given tomorrow's deadline, it won’t be possible to submit this genome on time. Sophie and I are awaiting Tyler's feedback for a final review, and we will submit it once that's completed 👍 . 

Thank you for your understanding.

Best, 

Lola

talioto:
Man, this one is not easy, but I assume you did some alignments to other Pipistrellus bats. Is the X colinear? Are there any Y's to align to. I'm not sure the piece labeled Y is the Y. I broke off the little bit that matches the X and placed in the gap in the middle of X.
SUPER_13: 53.6 to the end I would break off an unplace it in the chaff.

Here's a link to folder where I have the [pretextmap and a savestate](https://ccnag-my.sharepoint.com/:f:/g/personal/tyler_alioto_cnag_eu/EjNDttu3g59JgyStNX61EIkBVI6lI2m8Auh1pQU_XqLh7Q?e=ugzgW9).

ldemirdj:
Yes, it's a hard genome, good luck with this assembly. Sophie will answer your question. Thanks @talioto.

additive3:
My 2 cents on the Y chromosome.
What is currently annotated as Y looks to be just satellite.. perhaps centromeric, and looks like placement is to SUPER_3 (at a guess ~91.4Mb).
I agree that there is a small bit that clips off and is X centromere (also scaffold_56).

So Y... looking at the map, I would suggest that scaffold_32, scaffold_37, scaffold_58, scaffold_34 (in that order) are it.
 
Coverage is really not high enough and for another discussion.

SMangenot:
The Y chromosome was wrong, I made a mistake in my last card. 
I aligned the X chromosome against a reference genome and it now looks correct.
I followed @talioto's instructions for SUPER_13
I've organized scaffold_32, scaffold_37, scaffold_58, scaffold_34 (SUPER_22) to reconstruct the Y chromosome but the alignments against a reference don't seem conclusive.

additive3:
I wouldn't necessarily expect to see alignment between Y, esp. from different species.
While gene content is likely conseved, copy number, structure and composition are likely quite different.

talioto (approved):
Go ahead. Not much else to improve. Really need higher coverage of Hi-C from Genoscope to better scaffold and curate and spend less time doing it. The agreed on target is 50x coverage minimum. For bad libraries we go to 100x.

diegomics:
Hi @SMangerot, out of curiosity, do you know why HiC throughput was so low?

PR #90, opened 2024-10-15
Add Chaetopelma lymberakisi

auryjm:
## Assembly review request

- ToLID: qqChaLymb1
- Species: Chaetopelma lymberakisi
- Project: ERGA-BGE
- Affiliation: Genoscope


auryjm:
Please note that for this genome, the expected size was 1.2 Gb, but we ended up with an assembly of 5 Gb. This explains the relatively low coverage of the PacBio data (20x). However, we stopped at this point since the contig N50 was well above 1 Mb. As for the Hi-C, we sequenced two libraries, but again, the coverage is low.

tbrown91:
Ok

tbrown91:
Hi @auryjm thank you for submitting the assembly. To my eye it is looking in pretty good shape, even with the limitations of the Hi-C library. My initial reaction was all very good apart from maybe the last ~30Mb of chromosome 28, which looks a bit off, but I can't figure out where else it should go, unless it is a separate chromosome. There are also some horrible repeat blocks which are causing some mayhem, but I think they look ok.

Let's wait for a proper reviewer, but hopefully a quick turnaround.

gitcruz:
Yes

gitcruz:
Hi @auryjm,

I think you've done an excellent job for a 5.2 Gb genome and the poor density of hic data.

My suggestions are:

1. Break the super 28 in three pieces based in the contacts there are 3 different chromosome territories that can be broken at gaps. In my opinion these are three different smaller chromosomes 

2. First section of super 32 should be unplaced. Look at the oof-diagonal contacts. Strong with the rest with the chaff and mild with the rest of the superscaffold.


Please let me know if you agree.

Thanks
Fernando

tbrown91:
thanks @gitcruz 

I _think_ I'm happy with the join at the start of scaffold 28 as it is, but not the one at the end. @auryjm do you have any strong opinions or other supporting data? I couldn't really find much information on the karyotype of this species or related ones (seems to range from 18 to >60 chromosomes for this family)

Was the start of 32 joined automatically, or did you join it yourself? I guess this must belong somewhere, but I struggle with where.

gitcruz:
Hi @tbrown91 

28: Yes, perhaps the level of contact between the start (0-17Mb) and next block upo to 85Mb is "slightly higher than with other chromosomes". But the last bit is a bit orphan of contacts with the rest...to me has similar level of contacts with any other superscaffold.

Let's read @auryjm opinion.

Cheers,
F

auryjm:
Hi,

Thank you both for the review !

Good catch on scaffold_28. Indeed, it seems like there was a mistake during the chromosome painting process. I propose to keep the first two parts in the same chromosome and create a new chromosome with the end of the current scaffold_28. 

Regarding scaffold_32, it aligns with many other chromosomes, so I suggest we exclude this scaffold from the chromosome set. 

If this works for both of you, I will go ahead and generate a new EAR report !

JM

tbrown91:
Sounds good to me!

Thanks @auryjm and @gitcruz

gitcruz:
Great, once you have done these changes please generate a new EAR report.
 @diegomics would it be enough if he just provide a link to the new
save_state in the notes field?
remapping the hic for these few changes does not make sense.
Best
F.

auryjm:
Thank you !

I agree that this involves a lot of calculations for a few changes, but it will also affect the overall metrics, and possibly BUSCO, Merqury, and in our case, the number of chromosomes. It might be cleaner to generate a new EAR each time.

gitcruz:
Yes @auryjm,

But I think @tbrown91 and @diegomics will agree that is not necessary to remap the hic and rerun TAPAs...and I doubt the BUSCO or Merqury will significantly change...

By the way by " scaffold_32, it aligns with many other chromosomes" you mean that the first block of ~8-9Mb shows hic contacts with many other and you plan to leave this as unplaced. Did I get you right?

Thanks

tbrown91:
Yes, please generate a new report.

The statistics and plots should match what is published on ENA and in this case it is important that the TAPAS map has the correct scaffolds included. 

The BUSCO and merqury runs should not take too long to run. If you would like to included the newly-painted pretext map and image without remapping, I would be happy with that, but again this should run over the weekend.

As long as this genome is submitted next week, I will be happy

Thank you both again

auryjm:
Right, Fernando, I'll leave these small scaffolds as unplaced.

Calculations are running and should be done by next week!

tbrown91:
Thanks @auryjm 

have a nice weekend!

tbrown91:
Thanks @auryjm looking good!

@gitcruz if you are happy with the assembly, could you please "approve" the request and then we can merge

Thank you both

gitcruz:
Hi all,

For my the contact map looks good now. The only comment is that there is some naming problem with one of the unplaced scaffolds, _qqChaLym1_scaffold_34_unloc_1_ shoul be  scaffold_34 instead. am I right @auryjm ? If so just be sure you rename it before the ENA upload. I will approve asap if this is the case.

Thanks,
Fernando

auryjm:
Good catch, @gitcruz. It should be qqChaLymb1_31_unloc_1 as in the previous version of the assembly, possibly an issue during chromosome painting. I'll rename it to qqChaLymb1_31_unloc_1 without generating a new EAR report since it won't affect the information in the report. Does that sound good to both of you?

tbrown91:
Yes, thank you @auryjm

gitcruz:
Yes, sounds good. It's enough if you fix it before the ENA upload.

Congratulations!

gitcruz (approved):

PR #109, opened 2024-10-30
Update get_EAR_reviewer.py

diegomics:
Fix current implementation that is prioritizing older reviews before considering the number of total reviews


diegomics:
@arash77 I think this will fix the issue that I found the other day, do you agree? I'm also looking at the justification messages below the table...

arash77:
Yes, this will lead to better ordering by checking Total Reviews before Parsed Last Review.

diegomics:
Great, thanks for look at it. I'll merge it :)

PR #118, opened 2024-11-15
daArnMont1

DomAbsolon:
## Assembly review request

- ToLID: daArnMont1
- Species: Arnica montana
- Project: ERGA-BGE
- Affiliation: Sanger


diegomics:
Hi all, sorry the bot shold not call Jo here. I will fix this. @tbrown91 I'm appointing you and I'll try to fix the script asap!

diegomics:
OK

andar27:
Yes

tbrown91:
Hi @DomAbsolon 

Thanks for submitting the EAR and I don't envy you having to work with this sample. I don't know how much more time and money it would take to produce another Hi-C library. This one is really a challenge to work with., baybe we can get it into an _ok_ shape for submission as is. 

I have found the following areas of interest from my first look through. @andar27 please look through this as well soon so that Dom can get a thorough review of the assembly.

What do you think about cutting and flipping the first 1.3Mb of SUPER_4 and the last ~8.6 Mb of SUPER_7? I know there are no gaps here, but it would at least place the telomeres at the end and the Hi-C signal is not really that informative in either direction.

There are also some areas which look like haplotypic duplication to me:

Hapdup between SUPER_1-51.7Mb-52.7Mb and SUPER_19-23.8Mb-24.5Mb
SUPER_14 hapdup @ 36Mb
SUPER_19 remove some of these duplicate sequences around 17.7Mb? I'm not overly sure about this area. The coverage does drop to half in a number of these copies. Maybe just remove the contig from 17.687Mb-18.717Mb?

It looks like some of the smaller scaffolds are duplicates that could be removed to help with the genome-wide BUSCO scores - e.g. HAP1_SCAFFOLD_55, 56 & 61 (or 86, 93, 96, 145 & 229)

DomAbsolon:
Hi Tom, new HiC has already been ordered so I think this could get pushed through and a new and improved version can be done as and when we get that data

andar27:
Hi @DomAbsolon,
I went through the EAR and the contact map, I think you have done a really great job given the low-density of contacts for your map.

The reason being that due to the lack of signal it is quite difficult to determine with certainty the correct the assembly path. i think some additional HIC would help to resolve the regions of ambiguity.

Perhaps minimap2 alignments to other species in the genus would help sort some of this out ?

For SUPER_7 (telomere inversion) : this is as bit less certain but HIC seems to suggest leaving it that way !
SUPER_7

Adama

DomAbsolon:
Hi all, thanks for your comments. as we have ordered more HiC data generation we are inclined to pause operations on this assembly until that comes through. Under normal circumstances we wouldn't have pushed this out for release in the state it was in anyway, we tried to rush it through for the project deadline which was ultimately missed anyway so we feel better to wait now. I hope that is agreeable with you guys!

tbrown91:
@erga-ear-bot CLEAR

PR #127, opened 2024-12-12
xgPatRust EAR

fcamara7:
## Assembly review request

- ToLID: xgPatRust1
- Species:  Patella rustica
- Project: ERGA-BGE
- Affiliation: CNAG


EmilieTeo:
Yes

EmilieTeo:
Hi @fcamara7 

This assembly looks really good !
The only thing I would suggest would be to move scaffold_28 to SUPER_1. 
It looks like it could be placed at the beginning of SUPER_1, but being unsure of its orientation maybe it would be best to leave it as unloc. 
Except from this, I don't see anything else that would require to be changed

Best regards, 
Emilie

fcamara7:
> Thanks for the detailed curation notes @fcamara7, very helpful! I agree with @EmilieTeo, scaffold_28 shows more signal with SUPER_1. Is there a reason why you guys decided to not placing it as unloc?

Thanks to both @EmilieTeo and @diegomics for the quick review and comments. In fact you are right, I think I just missed scaffold_28 (not sure why..). Decided to put it as an unloc in SUPER_1 as I agree the segment is too small to orient properly..I have made a save state reflecting that modification as well as re-painting the scaffolds. I will proceed with re-mapping the HiC if you think that this is fine. I will not re-run blobtools but will regenerate the stats.

diegomics:
Remaping is not necessary yet. @EmilieTeo would you like to take another look to see if you spot something that requires a correction or clarification before approval?

EmilieTeo:
I checked again with the savestate, everything looks good to me

fcamara7:
> Remaping is not necessary yet. @EmilieTeo would you like to take another look to see if you spot something that requires a correction or clarification before approval?

@diegomics let me know whether I should proceed with the remapping now please, thanks!

diegomics:
> I checked again with the savestate, everything looks good to me

Good, then please approve it :)

diegomics:
Hi @fcamara7, no need to do a remapping or other reanalysis, the results will be the same in this case. I couldn't spot anything, so the EAR is good. Wait for @EmilieTeo approval

EmilieTeo (approved):

PR #130, opened 2025-01-17
Refactor EARBotReviewer to handle merged but unreviewed reports

arash77:
This is an update for #54
Enhance the EARBotReviewer to support the addition of YAML files for merged reports that have not been reviewed, streamlining the process of handling such reports.


diegomics (approved):
Awesome, let's check it!

PR #135, opened 2025-01-24
Update EARBotReviewer to not raise any error on EAR PDF modification

arash77:
This is an update for #54
Change the behavior of EARBotReviewer to print a message instead of raising an exception when an EAR PDF file is modified.


diegomics (approved):
Nice!

PR #138, opened 2025-01-31
Armeria pungens 

CaroB-M:
## Assembly review request

- ToLID: dcArmPung1
- Species: Armeria pungens
- Project: ERGA-BGE
- Affiliation: Genoscope


tbrown91:
ok

jesgomez:
Yes

tbrown91:
Hi @CaroB-M thanks for uploading the EAR!

I am a little bit concerned about how much appears to be missing from this assembly. Looking at the kmer spectra, there appear to be quite a significant number of single-copy kmers in the reads that were not included in the final assembly, coupled with a lower-than-hoped BUSCO score.

It is stated in the notes that around 500Mb of sequence was flagged as haplotypic duplication and removed, but this does not match the pre- and post-curation statistics. Do you mean during the purge-dups step? I would suggest that perhaps there is some over-purging taking place and removing both genuine paralogs/duplications as well as single-copy sequences from the genome.

Do you have the BUSCO scores and kmer spectra of the assembly pre- and post- purging?

image

@jesgomez I have not had a look at the Hi-C map yet. Any other comments from you at this point?

jesgomez:
I agree it would be worth having the pre-purging stats. Also, maybe you can share the plots produced by purgedups? They will help to see if changing the purgedups cutoffs is needed.  I will take a look at the Hi-C map later this afternoon. Thanks

CaroB-M:
Hello,

this is tle plot produced by Purge_dups : 
image

The busco scores are : 
before purge_dups : C:96.4%[S:87.9%,D:8.5%],F:1.2%,M:2.4%,n:1614
after purge_dups: C:87.3%[S:82.3%,D:5.0%],F:1.6%,M:11.1%,n:1614

You're right that it seems that purge_dups over purged the assembly even if the graph seems correct. 
I'll try to reintegrate the contigs and get back to you with a new assembly version.

tbrown91:
Hi @CaroB-M I think you have added your pdf to the wrong location. Can you move it and make a new commit?

tbrown91:
Overall looking very good and kmer spectra looking much better. Some areas I've highlighted that I think require another look:

Potentially move contig 308.75Mb-309.2Mb to 251.1Mb in reverse orientation. However this is clearly an area where the arrangement of the sequences differs between the two haplotypes.

Scaffold_18 potential haplotig to remove? Strong synteny with end of chr_8

Scaffold_25 potential haplotig - syntenic to chr_4 193Mb-194.8Mb

Scaffold_20 & 22 likely haplotigs from chr1 @ ~18Mb and ~40Mb

CaroB-M:
Sorry @tbrown91, I've moved the file to the right place

CaroB-M:
Hi @tbrown91 thanks for your suggestions! The assembly is better thanks to the addition of contigs which didn't have to be purged by purge_dups. I'm waiting for @jesgomez's suggestions and I'll try to be quick to allow the submission of this genome before the end of the month.

jesgomez:
Hi @CaroB-M  and @tbrown91! I agree that stats and merqury plots are looking much better now! Also what I managed to see from the ContactMap, it looks great. However, I would like to take a deeper look but I'm currently having issues with PretextView on my laptop, for some reason, it is going damn slow. Let me try to fix this and come back to you later today or tomorrow. Sorry for the delay!!

tbrown91:
have you tried the new version? Diego swears it is faster

jesgomez:
I can't even install it right now, it is apparently much faster, though

jesgomez:
Ok, I managed to review it now. I agree with Tom's suggestions and additionally I would suggest making scaffold 24 an unloc to chrom 8.

CaroB-M:
Hello,
thanks!
I agree with your suggestions. But for this one "Potentially move contig 308.75Mb-309.2Mb to 251.1Mb in reverse orientation. However this is clearly an area where the arrangement of the sequences differs between the two haplotypes.", I m not confident in moving this contig.
If it is OK for you, I will generate the new fasta and the new EAR file.

tbrown91:
sounds good!

CaroB-M:
Hello, I generated the new EAR and the new map with the modifications you proposed

jesgomez (approved):
After having arranged our suggesstions, I think this is now good for submission to the ENA. Thanks @CaroB-M!

tbrown91:
Thank you both!

PR #141, opened 2025-02-05
Add Lewinskya acuminata

ldemirdj:
## Assembly review request

- ToLID: cbLewAcum8
- Species: Lewinskya acuminata
- Project: ERGA-BGE
- Affiliation: Genoscope


diegomics:
Ok

talioto:
Sorry, I missed the request. I can still review it if needed. Looks fairly straightforward. Would be nice to see the usual extensions (gaps, telomeres, coverage, etc.). you may want to resubmit.

ldemirdj:
Hello, I agree with Tyler, I just saw that the usual extensions were missing, I will update the file.

JoannaCollins:
Hi @talioto

thanks for your message and comments re the tracks. I am happy to take on the review as its been a while since I did one I hope thats ok.

 @ldemirdj thanks for updating the map with the tracks :-) I will check it now.

JoannaCollins (changes requested):
Hi Lola,

I have checked this and found the following:

cbLewAcum8_1:
There is a large haplotypic inversion between ~3.79-4.36Mb. Typically we would make reference to this in our Genome notes which accompany our assembly submissions. You might want to put a note in your curator's notes? The region from the gap at ~37.24 to 37.42 is haplotypic duplication based on HiC affinity and coverage, this needs to be removed.

cbLewAcum8_2:
Looks like hap dup at ~ 28.59, I think the easiest way to deal with it is to remove the sequence between the 2 gaps. Looks like hap dup @ 33.14-33.18, this needs to be removed. This is quite difficult to see but I think the region between the gaps @ ~37.16 and 37.28 needs to flip. The region to the left of this between the gaps looks like it might be partially hap dup but its to difficult to see exactly what is going on here so i think you will have to leave as it is.

cbLewAcum8_4:
Looks like hap dup @ ~37.93-37.97 based on HiC cov and PB coverage, I think this region needs removing.

cbLewAcum8_5:
Region between gaps @ 11.21 and 11.35 needs to flip. From ~28.16-28.23 looks to be hap dup, i would remove this region.

cbLewAcum8_6:
It looks like hap dup @ ~29.85-29.87 i would remove this piece. Same again @ ~30.03-30.04 which i would also remove as hap dup.

I notice that you have the organelles in the primary assembly? I just wanted to note this as we submit these seperate from the main assembly so these should possibly be removed?

 
Please ask if you have any questions

 

Thanks very much, Jo

auryjm:
Hi  @JoannaCollins,

I wanted to follow up on your comment about organelles in the primary assembly. You mentioned that organelles should possibly be removed since they are submitted separately from the main assembly. We have always included the organelles with the primary assembly for both our BGE and ATLASea genome assemblies, and it appears that other centers follow the same practice. 

Could you clarify if there has been a change in this procedure? Are organelles now to be submitted independently? Could you provide some insight into the reasoning behind this change? 

Thank you for your clarification.

JoannaCollins:
Hi @auryjm and @ldemirdj 

Apologies for the confusion, there has been no change in procedure, this is totally my error.   We do not have the organelles included in our assembly during our curations and I had forgotten that they then get folded into the assembly as part of our automated post processing procedure before being uploaded to the ENA.

Leaving them in the assembly in the assembly is absolutely fine.

Best Wishes, Jo

ldemirdj:
Hi @JoannaCollins, 

Thanks for your suggestions, I totally agree with you. I made the changes and regenerated the EAR, it should be available on Monday.

Best, 

Lola

diegomics:
There is not much to add to Jo's great review :)
Maybe a possible hap dup on cbLewAcum8_4: 36.974M - 36.998M ? Not super sure, though

ldemirdj:
Wait, the EAR report is incorrect, a new one is coming.

JoannaCollins:
Hi @ldemirdj and @diegomics,

Sorry, I wasn't in work yesterday to reply to any comments. 

@diegomics I totally agree the region cbLewAcum8_4: 36.974M - 36.998M could easily be a hap dup based on coverage. Although its not very clear for me to be 100% sure based on the HiC data, its one of those slightly ambigous tricky ones. 

I think though that @ldemirdj has taken something out of the region though now and it looks ok as it is in the most up to date map :-).

diegomics:
Hi @JoannaCollins, if there is nothing else you would like to point out in the report, please approve the PR so we can close this one.
One question about the BUSCO numbers: Is the level of missingness typical for species in this class/order? @ldemirdj

JoannaCollins:
Hi @diegomics and @ldemirdj,

I was just speaking to Jo (Wood) about the BUSCO scores for this species and he suggested that the "missing" score for cbLewAcum8 is broadly in line with the "missing" score in our Bryophtye genomes which is averaging aorund 13%. He suggests this is more likely a Busco lineage issue rather than missing sequence. 

As such it looks like this is good to go and i will approve.

Best, Jo

diegomics:
Thank you Jo and congrats on a new one Lola!
Merging now

PR #143, opened 2025-02-06
add Avellara fistulosa (daAveFist1) report

zilov:
## Assembly review request

- ToLID: daAveFist1
- Species: Avellara fistulosa
- Project: ERGA-BGE
- Affiliation: Sanger


additive3:
@diegomics I should not oversee the QC of this genome

EmilieTeo:
Yes

diegomics:
@additive3 I’ll add this type of case to the selection system. @tbrown91 please supervise this review 🙂

EmilieTeo:
Hi @zilov ,
The region on chromosome 2 (245 - 275 Mbp) looks really odd. I don't know if it was scaffolded like so or if it was rearranged, but after moving some contigs around, it could look like a part of this region belongs to the second haplotype. It's harder to tell without having the alternate haplotype on the HiC map and I don't know if it was curated as well. 

Apart from this region, the rest of the map looks good to me.

image

 Emilie

zilov:
Hi, @tbrown91, @EmilieTeo

Thank you for such a precise feedback.

`You list the observed ploidy as 4. Where is this coming from?` - That's the error in the report file, will fix it in the next version.

`You say in the notes that there are two curated haplotypes. Are both included in the pretext map? The BUSCO scores suggest a single haplotype.` - By team guidelines, we are providing only a haploid map of the assembly for QC (most times hap1). All statistical metrics are calculated for this first haplotype.

- [x]  `Retained haplotig SUPER_1 around 94.1Mb` - agree, fixed
- [x]  `Likely duplicates here indicated by drop in coverage. SUPER_1 at ~319Mb` - agree, fixed
- [ ]  `Potential misassembly at the start of SUPER_2 at around 9Mb, or is this just due to the Hi-C signal?` I suggest that this is an artefact of HiC signal, no gaps around the region and normal hifi coverage. There is telomere signal in the beginning as well (blue track on [screenshot](https://github.com/user-attachments/assets/94e3c020-20c3-4118-b483-a35a8109418b)) and high repeat density on 9mb (green top track on [screenshot](https://github.com/user-attachments/assets/8ec98a86-1eb5-43ef-86c0-618e69c118c5)).
- [ ]  `I would suggest there are a lot of retained haplotigs in the region you highlighted in SUPER_2 from 247Mb-270Mb, e.g. 249.6Mb-250.4Mb, 254.5Mb-255.9Mb, 262.2Mb-263.7Mb.`
This pattern is common in regions with many repeats (probably rDNA) and isn't always indicative of haplotigs. Such a signal in repeats is an artifact of the Hi-C map and doesn't disappear when contigs are moved. In fact there was some retained scaffolds in the region, so I moved it to the the second haplotype. However, despite the fact that the region looks normal in high resolution, a "haplotig" pattern still appears on the map after remapping (see screenshots [before remapping](https://github.com/user-attachments/assets/923acbea-602b-4749-a49e-2270bc1af2b5), [after remapping](https://github.com/user-attachments/assets/5b5256ed-42fd-4e85-811e-7f8a87cea396)). I suggest leave is as it is or move all repeat sequence to Unloc of this chromosome. 
- [x]  `Potential retained haplotig in SUPER_5 from 78.5Mb-78.7Mb` - agree, fixed
- [ ]  `Potential retained haplotig in SUPER_7 from 30.8Mb-31.7Mb` - for that haplotig I see the same repetitive pattern on second haplotype ([hap1 placement](https://github.com/user-attachments/assets/9491efca-df83-44c2-91ee-7e1eec4214cb), [hap2 placement](https://github.com/user-attachments/assets/f45a5a6f-4aad-4de3-9694-99e363c4acf3)) so I suggest to leave in hap1. The thing is, that when I am working in two haplotype map I don’t see so much coverage drop in the region as on single hap1 map, gaps are in the same place for both haplotypes and that suggest that this contig should be on both haplotypes, since we want to have hap1 more completed I would leave it here.

Let me know what you think about non-checked cases. Once all checked, I will run HiC realignment, update fasta, chromosome files and report PDF file.

tbrown91:
Hi @zilov. Looking good.

I think I would be more comfortable if you moved the first quarter - or all - of the remapped region of SUPER_2 as you have it to the unlocs, then I'd be happy to sign this off.

@EmilieTeo any further thoughts?

EmilieTeo:
Hi @zilov, @tbrown91
Thank you for those precise information
I don't think I have anything to add, and I would also agree with moving the first part of the SUPER_2 region as unloc

tbrown91:
Hi @zilov
hopefully you have enough information to finish this assembly now.

Thank you @EmilieTeo for the comments

tbrown91:
Thanks @zilov 

@EmilieTeo can I ask you to look over the genome and if you are happy "approve" the review?

zilov:
Hi @EmilieTeo, @tbrown91,

Sorry for delay, it takes some time to re-align all reads and get updated assembly. I have updated PDF. 
Let me know if you need anything else to review :)

PR #145, opened 2025-02-11
Add Eunicella cavolini

ldemirdj:
## Assembly review request

- ToLID: jaEunCavo1
- Species: Eunicella cavolini
- Project: ERGA-BGE
- Affiliation: Genoscope


epaule:
Yes

epaule:
nice to see an octocorallia as part of BGE :-)

tbrown91:
Hi @ldemirdj 

I had a quick look through and found some funky areas. Do you have an indication of how to interpret the kmer spectra. Is this just an incredibly heterozygous species, hence the low kmer completeness? There also look to be a fair number of duplicate sequences in the assembly based on the merqury plot. Do you have the BUSCO scores before/after purging to give an indication of how successful this stage was?

The final ~2Mb of chr1 could be a haplotig. I moved it to 16.8Mb here. However going through the genome there are quite a lot of these situations that look like they could be biological based on read coverage.
image

Invert the first 4.5Mb of chr 2
image

Should the last ~600Mb of chr 10 be unloc?
image

chromosome 12 unlocs are potentially coming from the other haplotype - I kind find a nice place to put them in the scaffolds.

@epaule if you get a chance to look through as well this week, that would be great. We need to push through as many assemblies as possible within the next 2 weeks

epaule:
I will have to look at it a bit more, but I agree on declaring the end of _1 a hap and inverting the start of _2. If you cut _2 at 30.3, it could also join nicely onto _1 (inverted). _2 30.3-34 is also duplicated in _1 , but the coverage looks fine and there is no gap. _5 17.4-18.6 looks also weird.
It looks a bit like YAHS or HIFIASM smashed the haplotypes together in a not so nice way.
I will get some more opinions from the ASG experts and come back to you.

epaule:
we discussed the map today with colleagues and some things were:
* _11 is probably two chromosomes, as the telomer dot is not only at the ends, but also in the middle where the break in HiC signal also is located and suggests a split. There are some other chromosomes, where the telomer dot/location needs investigating.
* there are quite a few probable haplotypic regions, which are a bit difficult to see, as the coverage is very choppy. But the vote was for going by HiC signal, as long as there is a small dip in coverage 
(I will come up with a more detailed list of the things we spotted on Monday)

ldemirdj:
Hello everyone, 

I appreciate your feedback and will take all your comments into account. I will send you the updated EAR report shortly. 

Thank you!
Lola

epaule:
we came up with the below on Friday (i also attached a savestate). Looking at the BUSCO duplication rate, the hapdups from HiC might be not all be real hapdups, but they would need investigation.

chromosome 1:
333K-2.026M -> 3.080M -
0 - 333K hapdup
29.91-30.6M hapdup
85.8M-end hapdup

chromosome 2:
0-4.45M -
4.45-30.4M -
5.328-5.571 -
0-30.4M -> chromosome 1 @ 85.7M

chromosome 4:
51.1M-end hapdup

chromosome 5:
17.4-18.6M hapdup (probably more)

chromosome 7:
8.5-9.17M haplotypic region
12.228-12.433M haplotypic region

chromosome 10:
I would unloc 17M-end

chromosome 11:
needs splitting at 8.53M
16.25M-end are three hapdups

chromosome 12:
unloc 1 & 2 look like a hapdups

chromosome 14:
4.9-5.1M haplotypic region

[jaEunCavo1_hr_mh6_2.pretext.map.savestate_1.gz](https://github.com/user-attachments/files/18827588/jaEunCavo1_hr_mh6_2.pretext.map.savestate_1.gz)

ldemirdj:
Hello @epaule  and @tbrown91,

Thanks for your feedback. I've made the necessary modifications based on your comments:  

- I've checked the k-mer spectra, and it does seem to reflect the heterozygosity of the species. However, the plot appears slightly compressed, so I have attached the KAT plot below.
téléchargement1

- I agree that from Merqury's graph the assembly appears to contain a significant number of duplications, but identifying them accurately remains a challenge, as there are many similar areas throughout the genome.

Here are the **BUSCO scores before purging**: C: 95.7% [S: 90.6%, D: 5.1%], F: 2.7%, M: 1.6%, n: 255, and **after purging**: C: 87.9% [S: 85.9%, D: 2.0%], F: 2.4%, M: 9.7%, n: 255. Given the increase in missing genes, we decided to reintegrate contigs >1 Mb. You will find below the **KAT plot (post-purge_dups)**, where a black peak appears, suggesting that purge_dups may be removing too much. I have also included the cutoffs plot.
téléchargement3
téléchargement2

Initially, **208 regions** totaling **168 Mb** were identified as haplotypic duplications and removed. After reintegration:
    - **Cumulative size**: 372,980,833 → 517,457,888  
    - **Sequence number**: 95 → 127  
    - **N50 (L50)**: 12,728,091 (9) → 9,874,030 (15)  
This correction seems to better reflect the assembly's true complexity.

- I have applied the suggested modifications to **chromosomes 1, 2, 4, 5, 7, 10, 11, 12, and 14** based on the identified haplotypic regions and duplications. I am currently regenerating the EAR report so we can discuss it further.

Let me know if you need any additional details or adjustments.  

Best,  
Lola

tbrown91:
Hi @ldemirdj looking much better already. Hopefully this should be good to submit now. We'll have a look through the report and Hi-C map once you've generated them and hopefully sign-off soon

epaule:
I think it looks better this way, pretext-wise. But I think you still have some old numbers in the EAR (like the observed haploid number)?

ldemirdj:
Hi @epaule, 

These are the statistics provided by GOAT. However, they indicate that the number of chromosomes is between 6 and 16, depending on the ancestors' estimates.

Best, 

Lola

epaule (approved):
ah ... I see . My bad.

tbrown91:
Thank you both!

PR #157, opened 2025-02-18
Add Prunus ramburei genome assembly

auryjm:
## Assembly review request

- ToLID: drPruRamb1
- Species: Prunus ramburei
- Project: ERGA-BGE
- Affiliation: Genoscope

PR #161, opened 2025-02-21
Update Prunus ramburei EAR

auryjm:


diegomics:
Thanks @auryjm 
I reviewed the new EAR (this updates the very recent #157).
I can see the new EAR shows a very subtle size increase, gap size increase, and also one scaffold less (also visible in the TAPAs plot). Can you explain this? The Curator notes don't show anything new in that regard. Is this the result of integrating the unloc into Chr_1 pointed after the PR merging?

diegomics:
Great, merging now, thanks!

PR #170, opened 2025-03-27
xgPhoTurb1 EAR

gitcruz:
## Assembly review request

- ToLID: xgPhoTurb1
- Species: Phorcus turbinatus
- Project: ERGA-BGE
- Affiliation: CNAG


tbrown91:
ok

bistace:
Yes

bistace:
Hi @gitcruz!

in the stats and in the pretext map made with HiC from the sister species, I can see that there is only one gap. In the curation notes it is stated that the Hifiasm assembly was highly contiguous and if I understood correctly you organized the scaffolds based on the P. lineatus chromosomes. Did you have to do some manual reorganization of contigs or was the assembly already chromosome-scale (with the exception of SUPER_17)?

gitcruz:
Thanks for agreeing @bistace and @tbrown91,

Here we are submitting the xgPhorTurb1 for review, as detailed in the note we couldn't get hic data for it. However, we have strong evidences supporting the high quality and chromosome-level of the assembly.

Please let me know what you think.

Best,
Fernando

bistace:
Thank you for your answer @gitcruz!

The HiC data seems to confirm the inversions that we can see on the dotplot vs P. lineatus and the TIDK plot suggests that chromosome arms are well oriented. I only have a few suggestions about the smaller contigs:
  - ptg000044l seems to be a duplication of the end of SUPER_7 based on coverage and HiC pattern
  - ptg000572l has contacts with SUPER_17 so it may be marked as unloc
  - ptg000573l has contacts with SUPER_8 so it may be marked as unloc

Could you please look at that and see if you agree? Here is my [savestate](https://github.com/user-attachments/files/19487893/xgPhoTurb1_hfsm_pri_join_mt.decont_mq10.extensions.pretext.savestate_1.txt) so it is easier

gitcruz:
Thanks @bistace!

Excellent, these were very good catchs, specially the haplotig. I was personally reluctant to curate based on the sister species hic, but I agree on your edits. I think we can placed this 2 small contigs and remove 1 haplotig.

If @tbrown91  agrees, I can just produce a new fasta and upload it to ENA. I don't think there's much need to produce a new EAR report, right? I don't think this will change the stats that much...

 Best regards,
Fernando

gitcruz:
Hi @tbrown91,

At first I can look over the pretext map and see if there are drops in coverage at the inversions...let me check.

Best,
Fernando

tbrown91:
I wonder if the resolution is not good enough. Hopefully you still have the bam or bigwig files that you can look at in IGV. It should be very obvious if the coverage drops to zero or if there are lots of clipped reads

gitcruz:
I am looking into these rearrangements. Take into account that the precise alignments found have low identity <0.1 due to the  divergence of at least 15MYA between both species.

gitcruz:
If coverage would drop to zero, we will have a gap at the boundaries of the
inversions visible in the Contact map (actually a few compared to the small
ones found with the WGA) and also would see a dramatic drop on coverage at
the extension ONT_cov. I dont see that.

tbrown91:
HI @gitcruz actually, that's not my experience with hifiasm, which is a very "greedy" assembler and will try to continue contig paths through whatever means necessary. These misjoins are often fixed by scaffolders or curation, hence why I ask here as this cannot realistically be done. An example I came across here: https://github.com/chhylp123/hifiasm/issues/54

Is the resolution of pretext high enough to see such small-scale changes in coverage?

gitcruz:
Hi @tbrown91 

I have checked the ONT alignments at 6 inversions that are quite obvious in the pretext map (marked the flanks with the waypoints in the savestate: xgPhoTurb1_hfsm_pri_join_mt.decont_mq10.extensions.pretext.savestate_2_waypoints_rearrang ). I also selected the regions 5Mb around these waypoints and look at the coverage. 

These are large inversion that can be "fixed" using the sister species contact map, and everything looks great afterwards. However, the inspection of the ONT alignments using IGV shows long read alignments spanning these "waypoints" were the inversion starts or end. There are no zero coverage gaps. So, the oberved inversions in the contact map are actually "real rearrangements" or difference betweem _P. trurbinatus_ and _P. lineatus_

You can check this presentation summarizing the IGV inspection [here](https://ccnag-my.sharepoint.com/:b:/g/personal/tyler_alioto_cnag_eu/Edwbxwpj4IZNoliqTAGpVVsB2KafTI4WFMUZMOhUZ9MJJA?e=B4LOv4)

The pretext also shows small-scale rearrangements that seem to be real but are harder to fix, often because they are mixed with inversions. There is a strong contact between chrom18 and 16 (SUPER_18 and 14) that seems to be a subtelomeric interaction as is not supported by the WGA.

One more thing, we are using hifiasm 0.24.0 and from what you wrote in https://github.com/chhylp123/hifiasm/issues/54 from v0.15 onwards you haven't experience these problems...

Please let me know what you think now about the inversions. Although is a bit heavy, I'll be happy to provide the bam with these slices of the alignment.

Thanks,
Fernando

tbrown91:
Hi @gitcruz thanks for looking into these regions. That's looking really good. Given the lack of Hi-C from this species, I wanted to be as careful as possible.

Regarding the report, can you please update any statistics that change based on @bistace's comments. I think there is just one sequence to remove as a duplicate.

Happy to move to the upload and then once you have a new pdf, @bistace can "approve" the review and we can merge.

Thank you both!

gitcruz:
Hi @tbrown91, 

Yes, it makes sense to be thorough reviewing the assembly in absence of HiC. I really appreciate the feedback obtained during the EAR reviews. We always end up a much better assembly version.

I will produce a new fasta including @bistace edits, recompute the stats, and upload the new EAR asap.

Thanks to both of you!
Fernando

gitcruz:
Hi,

I'm waiting for @bistace approval and @tbrown91 to merge with main branch. I'll upload to ENA asap.

Thanks

bistace (approved):
Hi @gitcruz,

thank you for applying the changes, I think that everything looks great now. Congrats for this new genome!

gitcruz:
Thanks @bistace and @tbrown91!

PR #173, opened 2025-04-03
update ear diadema setosum

Arno75:


diegomics:
Hi @Arno75
I reviewed the new EAR (this updates #72).
Please explain what's different between the previous EAR and this one, I see much more going on than a mitochondrial sequence added.

diegomics:
> Hi @Arno75 I reviewed the new EAR (this updates #72). Please explain what's different between the previous EAR and this one, I see much more going on than a mitochondrial sequence added.

@Arno75 @ldemirdj can I get a response here please?

ldemirdj:
Hi @diegomics , I’ve asked Arnaud to reply directly — I hope we can finalize this quickly. Sorry for the delay, and thanks again!

Arno75:
Hello @diegomics  ,

 I only added the mitochondria sequence in chrom_list file, but it seems that I had problems with my first ear file. I have to regenerate it , to check the differences.

Arnaud

diegomics:
Ok, please confirm the differences. The previous EAR contained 98 scaffolds while the new one is showing 102. If only the mito was added the number shuld be 99. Also, the new total asm lenght is shorter than the previous one, should be the opposite (the difference is 158,146 bp). If the curation notes are not updated in the new EAR (which is the preferred), please state here what changed between the old and the new PDF, it's important for traceability.

diegomics:
All good. Please push the EAR_2 as the new eeDiaSeto1_EAR.pdf in the PR so I can merge it

PR #174, opened 2025-04-25
EAR upload of ihGraItal9

epaule:
## Assembly review request

- ToLID: ihGraItal9
- Species: Graphosoma italicum
- Project: ERGA-BGE
- Affiliation: Sanger


diegomics:
ok

diegomics:
Hi @epaule please check the PDF, some BUSCO metrics are missing

epaule:
@diegomics sorry, we will correct that.

bistace:
Yes

bistace:
Hi @epaule and thank you for submitting this genome!

Globally I think that the genome and the contact map look great. My only concern is with two small regions on the SUPER_1 (99.1Mb -> 100.3Mb) and SUPER_X (183Mb -> 189.5Mb) scaffolds. Indeed, there is a big coverage drop and a discontinuity in the Hi-C pattern, which could indicate a misplaced contig. Could you please check these two regions?

epaule:
Hi @bistace ,

for the SUPER_1 bit, you can see some comments in the "Curator notes" section.

The SUPER_1 position is the breakpoint of an haplotypic inversion, so hap2 has a large inversion and that allele is exactly where it breaks. Also hap2 has a shorter version of it. There is also lots more shrapnel with the same repeat sequence in the assembly. While I don't know if the repeat number of the allele is correct, I think it is as best as we can do, as it also got sequenced through (the contig covers it completely and more). So I would leave it as is.

On the region on SUPER_X, I can only spot a repeat with slightly dodgy coverage and two gaps, but I would also leave it as is, as i cannot establish how much of the repeat sequence in the shrapnel would go, as both gaps are internal (and the repeat is not unique to X).

diegomics:
Thanks for the explanation about the signal drop in SUPER_1 @99,067 Kbp. I'm happy with it. Also, I agree with the decision related to the repetitive region in SUPER_X that is kinda flanked by gaps @ 188,75 and 189,39 Kbp. Honestly, I don’t see anything else to point here, the EAR looks good.

bistace (approved):
Thank you for addressing my comments! I agree with Diego and think that this assembly is good to go, congratulations for this new genome!

diegomics:
Thanks @bistace and @epaule, congrants!

PR #182, opened 2025-06-18
EAR upload for qlOrcRanz

msozzoni:
## Assembly review request

- ToLID: qlOrcRanz
- Species: Orchesella ranzii
- Project: ERGA-BGE
- Affiliation: UNIFI


ldemirdj:
Hello @msozzoni, 

Thank you for sharing this genome assembly for review. I’ve gone through the provided EAR and would like to share some initial questions and observations:

- Could you clarify the reasoning behind manually curating both haplotypes instead of generating a haploid assembly, which is typically simpler and more commonly used for this type of analysis? Unfortunately, I was not able to open the Pretext contact map for haplotype 1.
- Could you please confirm the Hi-C and HiFi coverage? Also, was the sequencing performed with PacBio or ONT?
- On the post-curation contamination plots, I noticed that, in addition to "Arthropoda" and "N/A", there are other taxonomic groups present. Why were these retained instead of being removed from the initial assembly?
- I briefly checked the contact maps and noticed that the last scaffolds are all tagged. For some of them, I agree with the assignment to scaffolds, but I would have left most of them unplaced, especially those that show no clear contacts with a particular scaffold. There are also scaffolds that show no contacts at all — I would have tagged those as "Contaminant".
- To complete the curation review, I would need the usual additional tracks (coverage, telomeres, repeats, etc.).

Thank you in advance for your feedback.

Best regards,

Lola

tbrown91:
Hmmm... I think you need to check how the coverage track was made. When I looked at the hap1, the coverage is decreasing across SUPER_1 from ~130X down to about 20X and then is only 1/2X for the other scaffolds. Have you somehow sorted the values or something?

msozzoni:
Hi @tbrown91, 

I will check how the coverage track was made for these maps
I am fairly sure the map was not sorted 
I am now working on the new maps. Once finished, should I present only one or both of them for the re-review?

tbrown91:
Hi @msozzoni I think this is looking pretty good. Just a note to say that the coverage track looks to have the same issue again that the coverage is very high at the start of the genome and then drops to zero.

Once @ldemirdj is happy and Approves the Pull Request, I think you will be able to upload this soon

ldemirdj (approved):

tbrown91:
Thanks @ldemirdj !

PR #184, opened 2025-06-18
Add ilHipMade1 EAR

jesgomez:
## Assembly review request

- ToLID: ilHipMade1
- Species: Hipparchia maderensis
- Project: ERGA-BGE
- Affiliation: CNAG


diegomics:
ok

CaroB-M:
hi @jesgomez! This assembly looks very good! I have nothing special to say. 
Maybe the unlocs of the super 11,14,15 et 16 of the hap2 look like haplotigs.

diegomics:
hi @jesgomez
Congrats on the assembly, good metrics, looks good! Nothing caught my attention on hap1, maybe SUPER_22_unloc_1 in the same direction as @CaroB-M coments? But I'm not super sure. Maybe because I'm missing the repeats track... Anyway, please let us know what you think :)

jesgomez:
Dear @CaroB-M  and @diegomics, thanks a lot for your comments. I've been checking the sequences you mention and I am reluctant to tag them as haps, since in this case I'd have to move them to the other haplotype and the pattern there is very similar, actually those sequences already seem included. In fact, I'm more inclined to think they are repetive sequences (the map is done with mq0, rather than haplotypes). For example, see attached both SUPER_14: 
Screenshot 2025-06-23 at 11 19 07
Screenshot 2025-06-23 at 11 19 22

The ones in SUPER_11,15 and 16 could seem to go in the middle of the chromosome, but I have no GAPS in hap1, so I'd also not move them there. Same goes for SUPER_22 in hap1, the chromosome is also contigged in hap2 so I don't see the point in moving it. Do you agree? Having both haplotypes almost T2T is quite new for me, so maybe I'm not treating the sequences properly.

diegomics:
Good reply, done from my side. Keep in mind that Hifiasm may be a little greedy sometimes, and cuts may be necessary despite the absence of gaps :)

CaroB-M:
hello, OK for me too. I agree that the repeat track is more in the sens of unloc. OK to leave them as they are.

CaroB-M (approved):

CaroB-M:
done! Congratulations @jesgomez

diegomics:
Awesome, thanks both!

PR #188, opened 2025-07-15
icCryBari1 EAR

ilenia6:
## Assembly review request

- ToLID: icCryBari1
- Species: Cryptocephalus Barii
- Project: ERGA-Pilot
- Affiliation: UNIBA


SarahPelan:
Hi,
I'm having trouble accessing the curated hic map - when I download and click on the link I get directed to a login to ERGA Pilot Nextcloud which I've not seen before..
Any help greatly appreciated!
Sarah
Screenshot 2025-07-15 at 14 45 08

diegomics:
Hi @SarahPelan we are solving this, just a sec

ilenia6:
Hi @SarahPelan,
here's the link to access the curated contact map:

https://ergapilot.bsc.es/index.php/s/qQJRCYbagd2fZoX

Thank you!

SarahPelan:
Hi @ilenia6 
I have checked this and think that:

SUPER_1:
51.5-end looks like it needs to flip

SUPER_2:
6.65-6.78 haplotypic duplication remaining

SUPER_4:
40.4-end looks a bit better flipped

SUPER_8:
Scaffold_40 can join in @ ~16.1
18.0-18.1 looks like small hap duplication
I think that the order and orientation from ~26.0 may need to be changed:
flip ~29.1-32.2 join this flipped piece in at ~26.0
SUPER_8_unloc_2 can join in at ~29.0
I would still leave SUPER_8_unloc_1 as unloc

(Let me know if this is unclear)

SUPER_13:
0-428 haplotypic duplication
428-1.6 moves and joins in at ~4.3 but in opposite orientation
~10.6 to the end needs flipping, may need some further rearrangement within this section but see how it looks after you flip it and go from there.

(Let me know if this is unclear)

SUPER_14:
11.71-end looks better flipped

I would also incorporate some more of the shrapnel:
scaffold_27 to lh end of super_14
scaffold_29 belongs to super_8 but maybe as unloc
scaffold_25 - I would add this as X_unloc

Thanks,
Sarah

SarahPelan:
@ilenia6 and @tbrown91 I will look at this one again today

SarahPelan:
I have looked through this again - it looks much better, there a couple of things that I think need to change:
SUPER_8:
Regions like this are quite tricky and I know it's already been rearranged once, sometimes I can't quite see what is going on until I have the new map, I would suggest the following:
small piece from ~32.2-32.3Mb looks like it moves to ~29.3Mb, then move the piece from ~32.3-33.9 next to this (both flip)
I would flip from ~26.1-26.9Mb

SUPER_13:
flip from ~11.2-12.7Mb

SUPER_14
0-~526Kbp looks like it needs to flip

SUPER_X:
37.7-end I agree with the comment above that it either should be unloc, or joins to the start
Thanks,
Sarah

tbrown91:
Hi @ilenia6 I would suggest removing the .[2-3] from the file and assembly name. The version number should only be updated if this is an update to a published assembly (i.e. already on ENA).

SarahPelan:
@ilenia6 I'm getting an error when I try and open this:
Screenshot 2025-07-31 at 10 18 23

tbrown91:
Me too, although [this one](https://github.com/ERGA-consortium/EARs/blob/f902362ef46c277d44e523c16adc8e6bd71e8be9/Assembly_Reports/Cryptocephalus_barii/icCryBari1/icCryBari1.3_EAR.pdf) from before the rename is working for me

PR #190, opened 2025-07-15
bScoRus1 EAR

msozzoni:
## Assembly review request

- ToLID: bScoRus1
- Species: Scolopax rusticola
- Project: ERGA-BGE
- Affiliation: UNIFI


msozzoni:
Sorry for the mismatched file, I will update the EAR. The script microfinder was already used, but I will upload the fasta on drive and add the link here.

epaule:
not sure about the BGE SOPs, but would we want to orient and name the chromosomes based on GCA_964304615.2 ?

tommathers:
Hi @auryjm. We have found that if using hifiasm with pacbio + ONT given with the --ul flag then the micros still have missing content because hifiasm uses the pacbio to build the graph. If you have enough ONT data then I would recommend running the assembly again giving just ONT (with the --ont flag set (if you have HIC fromt he same sample you can also give the HIC data to get a phased assembly.

For the above to work you need to give it raw ONT data base called with dorado in super accurate mode (not eg. herro corrected reads). How much ONT coverage do you have? To increase coverage you can include shorter ONT reads for the de novo assembly eg. use reads longer than 10kb rather just filtering out for reads over 50kb or over 100kb.

epaule:
to get things going, I would propose to do the edits on this bScoRus1, submit it and then come back later, have another look and maybe do a ENA re-submission once work on the ONT only assembly has beed done.

msozzoni:
Great, I will produce the new pretext, BUSCO and gfastats scores

msozzoni:
Hi all, sorry for the double file and the delay. Here is the new EAR with the new pretext, BUSCO and gfastats with the recommended changes

tbrown91:
Ace! Thank you both @msozzoni @epaule

PR #193, opened 2025-07-22
drAlcFaer1

csantos-alvess:
## Assembly review request

- ToLID: drAlcFaer1
- Species: Alchemilla faeroensis
- Project: ERGA-BGE
- Affiliation: Sanger


tbrown91:
yes

tbrown91:
Hi @csantos-alvess thanks for sending in this monster. It's really making me question what words like haplotype and genome even mean any more.

Have you thought about how this is going to be submitted? From what I can see there are maybe about 80 groups of chromosomes that could be grouped together and called A/B/C/etc versions of each chromosome, but it's really difficult for me to say for sure. I am spending time at the moment just trying to group together ones that could belong together.

For example these seem to quite nicely group:

image

But here I have no idea:

image

I don't think it would be appropriate to submit this as a genome containing 211 distinctly named chromosomes and wherever we should, we should group them

csantos-alvess:
Hi @tbrown91, thank you for your comments and for your time reviewing it. Yes, this was a tough one. This species has a complex evolutionary history with many events of recombination and hybridisation along the time. Our main challenge with this one was when trying to separate the haplotypes and identify the ploidy, which we couldn't. During curation, we observed that the chromosomes could be gathered in groups of 3, 4, 5 and 6 chroms and sometimes we could see chroms that shared similarity with more than one group, as the one in the image I am sending and in your save state (SUPER_119). As the literature says, meiosis doesn't look to be viable for the species and reproduction is mostly clonal. Based on this, we think it would be very risky to try to separate the haplotypes or group the chromosomes when we are not sure if this is what really happens in nature. We discussed widely how would be the best way to submit it and we think the most proper way would be to submit the whole set of chromosomes, to avoid any errors when trying to group the them. Hope this makes sense.

Screenshot 2025-07-29 at 11 32 56

tbrown91:
Thanks for the explanation. 

Do you also think it is appropriate to name the chromosomes simply based on length? There is so much synteny, that it feels appropriate to group together as much as is possible at least in terms of the numbering.

tbrown91:
Alrighty then. If only we were reporting number of chromosomes produced in the project, we would be nearly finished with this one.

Thank you for the assembly, on to the next one!

PR #196, opened 2025-07-25
kaDisBerm1 EAR

ilenia6:
## Assembly review request

- ToLID: kaDisBerm
- Species: Distaplia bermudensis
- Project: ERGA-BGE
- Affiliation: UNIBA


ldemirdj:
Hi @ilenia6,

Thank you for submitting this genome. After reviewing the EAR report, I don’t have much to comment. Regarding the map, I repositioned an unlock on chromosome 5, and I believe there was also an inversion. Please find my `save.state` file attached : 
kaDisBerm_pre_review_normal pretext savestate_1

Best regards,

Lola

ilenia6:
@ldemirdj Thank you so much! Good to hear there aren't major changes to make. However, I don't know why but I'm not able to download the file you attached

diegomics:
Hi @ilenia6, nice assembly. I have a few small comments:
- I noticed that this was assembled with Necat. Out of curiosity, did you try hifiasm and it didn't work? Since the WGS and HiC data share the same sample accession I think was a good case for ont & hic modes assembly. Just wondering about it...
- SUPER_11 2.74 to 2.80 (gaps flanking) I think could be moved to the gap @ 2,09? Not super easy to tell at this resolution (HiGlass can help here)
- I see the point Lola raised about SUPER_5, and indeed looks like can be moved but I'm not super sure if it's better than the unloc though
- Something may be around 8.6 to 9.0 in SUPER_7, but again, not easy to tell at this resolution, HiGlass could also help there

There is a _real_ inversion towards the end of SUPER_6. It's handy to point these things in the curator notes :)

ilenia6:
Hi @diegomics , regarding the use of Necat as the assembler, the reasons behind this decision are related to the fact that in our first tests (we ran them almost three years ago) it turned out to produce the best results in terms of assembly continuity — other assemblers tended to produce very fragmented sequences. More recently, we’ve started to integrate the use of Hifiasm, especially now that it includes an option for working with ONT reads. Regarding the possibility of generating a phased assembly using HiC reads from the same individual as the ONT reads, I was aware of that as well. Unfortunately, for the species we work on, we were only able to produce Hi-C data quite recently (by then, the assemblies had already been generated and purged). So, even in the few cases where we did have Hi-C data from the same individual, we didn’t start the assembly process from scratch due to time constraints; we just continued working with the assemblies we already had.

Anyway, I updated the EAR report with the new contact map.

diegomics:
Ouukey! Still have the feeling that this assembly could be a little better with hifiasm, you know? Probably higher QV, which could bring higher BUSCO... And definitely there are some missing chunks here and there, that could also get assembled.
I think we can proceed with this one, but if you are itching about it, you can have the hifiasm run in just a few hours. If the hic data is from the same individual, you can run it in hic phasing mode.
Only things to add to the current one are:
SUPER_3 could be flipped from the beginning until the gap @ 1,3
Scaffold_51 could be linked to SUPER_4 as unloc
Please have a look @ldemirdj

ldemirdj:
Hi, 

I agree with you, that makes sense. Running hifiasm could indeed improve the assembly and address the missing chunks. The suggestions about flipping SUPER_3 and linking Scaffold_51 to SUPER_4 also sound good to me.

Thanks, 

Lola

ilenia6:
Hi @diegomics and @ldemirdj , I completely understand, and I also agree that the best-quality solution would be to assemble again the genome using Hifiasm and generate a phased assembly. Unfortunately, since I’m currently handling the manual curations of different genomes at the same time, fixing the map that was already created was the most convenient and quickest solution for me. I have uploaded the new report with your latest suggestions. Thank you again!

diegomics:
@ldemirdj please approve if there is nothing left from your side :)

PR #203, opened 2025-08-19
idEumHisp1 assembly review request

zilov:
## Assembly review request

- ToLID: idEumHisp1
- Species: Eumerus hispidus
- Project: ERGA-BGE
- Affiliation: Sanger


tbrown91:
@diegomics number of available reviewers has hit zero :)

tbrown91:
ok

zilov:
Hi @CaroB-M,

Thanks for the review, check my comments below :) 

I attach a save state with some suggestions.

- [x]  I added scaffolds 36,138 and 136 to chromosome Y between waypoints 3 and 4
    - [ ]  Agree
- [x]  I added scaffolds 163,139 and 98 to chromosome X between waypoints 5 and 6
    - [ ]  Agree
- [ ]  I added scaffolds 199 and 247 at the beginning of the chromosome X between waypoints 7 and 8
    - [ ]  I think for 247 there is not enough signal to place it in chromosome X - coverage is higher than X chr, and signal shows that it could be placed to SUPER_5 or SUPER_4 as well, so I would like to leave it in shrapnel

I made deux inversions :

- [ ]  Edit 3:SUPER_X[76.43 Mbp] to SUPER_Y[10.42 Mbp] nverted and moved to SUPER_X[76.40 Mbp]
    - [ ]  I could invert the region, but after remapping we will still have this inverted-like pattern, it is quite common for highly repetitive regions and it’s mostly mapping artefact. I left a note that the exact order and orientation of the contigs on chromosome X (75.4 - 79.2 Mbp) are unknown in genome note, which will be available after submission on ENA/NCBI. Alternatively I can tag all the contigs in the repeat as unloc. Let me know what works best.
- [ ]  Edit 4: SUPER_Y[10.40 Mbp] to SUPER_Y[3.12 Mbp] inverted and moved to SUPER_Y[10.42 Mbp]
    - [ ]  I'm not confident we have sufficient evidence to break contigs in region without gaps based on HiC signal only since the Y chromosome is highly repetitive and HiC signal could be funky because of it. Usually we rely on gaps before breaking scaffolds based on HiC signal. I see two possible solutions here:
    1. Leave it as it is and add comment that order and orientation of chrY contigs is uncertain. 
    2. Tag SUPER Y [8.0 -15.88 Mbp] as SUPER_Y because it has highier coverage, tag the rest of contigs as unloc.

- [x]  I added scaffolds 37, 71, 35, 44, 87 as unloc of the SUPER_1
    - [ ]  Agree
- [ ]  I added scaffold 124 at the position 96.731Mb of the SUPER_1 (waypoints 9-10)
    - [ ]  Not sure about that piece, it has haplotig like antidiagonal signal, lower coverage and has decent signal to SUPER_4 and SUPER_5 repeats. I’d like to leave it in shrapnel.
- [ ]  I have a doubt if the portion between waypoints 1 and 2 is not an haplotypic region. The coverage is lower than on the rest of the chromosome.
    - [ ]  Looking on merged haplotypes high-resolution map there is the same pattern on hap2 (see screenshot). Lower coverage is usual for centromeric regions, we have two copies of the region on both haplotypes and contigs are quite large, so I’d like to leave it as it is.
    
image_repeat

 
    

- [x]  I added scaffolds 33 and 34 to the SUPER_2 between waypoints 11 and 12
    - [ ]  Agree
- [x]  I added scaffold 140 as unloc of the SUPER_2
    - [ ]  Agree
- [x]  I added scaffold 30 to the SUPER_4 between waypoints 13 and 14
    - [ ]  Agree
- [x]  I added scaffold 132 to the SUPER_5 between waypoints 15 and 16
    - [ ]  Agree
- [x]  I added scaffolds 109,39 to the SUPER_5 between waypoints 17 and 18
    - [ ]  Agree
    
 
 Have a good weekend!

tbrown91:
Thanks @zilov looking good!

@CaroB-M could you please have a look through the assembly and let us know what you think

CaroB-M:
Hi @zilov! Thanks for the new map. Everything looks fine to me. Congratulations!

CaroB-M (approved):

tbrown91:
Thank you both!

PR #214, opened 2025-09-02
icGraMeto1 EAR

fcamara7:
## Assembly review request

- ToLID: icGraMeto1
- Species: Graciliella metohijensis
- Project: ERGA-BGE
- Affiliation: CNAG


additive3:
OK

diegomics:
Sorry for all the noise guys, the system should be working nicely again

fcamara7:
> Hi @fcamara7 and @additive3,
> 
> I see a lot of patterns that look like allelic duplications, with a gap in the middle but without the coverage drop that is normally associated with it. Like this one (coverage in red, gap in black): image What is your stance on this, would you tag one half as haplotig or would you leave it alone?

Thanks @bistace and sorry for the delay. We have been quite busy here these days.. I would normally tag this as an haplotig...but maybe @additive3 has a different take on it? I would not be surprised there are a few of these as this assembly is from a pool of two individuals. Given that we actually thought the assembly was quite decent, but I would have certainly missed some of the haplotigs. I am currently trying to make some changes to another (more difficult) cave beetle which happens to have a lot more haplotigs (pool of three individuals in this case). Best, Francisco

additive3:
Hi @fcamara7 

Sorry late to this review and agree with what @bistace suggested 
looking at the revision I think the assembly looks really good now with respect to the retained haptigs.
Again given a pool of 2 individuals this came together surprisingly well.
I think correct desision to not call the X_unloc_1 scaffold as belonging to X as I dont believe it is part of X... I wonder if potentially Y (although I wouldn't name it without certainty)

SUPER_7 the very end contig sequence is actually the start of the chromosome.
SUPER_11 I'd suggest given contact signal that scaffold_12 (198), 74, 63 and 22 all make up the short arm of that chrm, but equally understand any reluctance to place.

As with previous pool genome the plody for this should be 2 not 4 :)

fcamara7:
> Hi @fcamara7
> 
> Sorry late to this review and agree with what @bistace suggested looking at the revision I think the assembly looks really good now with respect to the retained haptigs. Again given a pool of 2 individuals this came together surprisingly well. I think correct desision to not call the X_unloc_1 scaffold as belonging to X as I dont believe it is part of X... I wonder if potentially Y (although I wouldn't name it without certainty)
> 
> SUPER_7 the very end contig sequence is actually the start of the chromosome. SUPER_11 I'd suggest given contact signal that scaffold_12 (198), 74, 63 and 22 all make up the short arm of that chrm, but equally understand any reluctance to place.
> 
> As with previous pool genome the plody for this should be 2 not 4 :)

Thanks @additive3 !, will take a look and make the suggested changes. With regard to the Y possibility perhaps I will refrain from assigning X_unloc_1 without further evidence(?).  As for the ploidy I think it should be "fixed" in the latest EAR report version!

PR #215, opened 2025-09-02
qqEusFeti1 EAR

gitcruz:
## Assembly review request

- ToLID: qqEusFeti1
- Species: Euscorpius feti
- Project: ERGA-BGE
- Affiliation: CNAG


andar27:
yes

gitcruz:
Hi @diegomics 

See my answers below.

> This one is a big boy, cool! It looks nice @gitcruz, let us know what you think about @andar27 ’s comments. 

Sure, it's big! I agree with @andar27 

I don't have much to add, I would need a repeats coverage track to go finer here. 

Well, we normally go with the coverage and multimappings (mq0 pretext)

The scaffold_136 mentioned in the curator notes is the Ascomycota showing in the blobplot?

Yes, exactly is scaffold_136.

 The BUSCO dups are a bit high, what can you say about that?
 
I would say, given that merqury found just 0.55% false duplicates and the genomesize is 4.37Gb, they might correspond to real duplicates in the 2-copy part of the spectra (blue line in the merqury plots).

Please let me know what you think about my previous post (I wrote it without reading yours first).

Thanks!
Nando

diegomics:
Agree, I'm happy. Please @andar27 approve so we can merge this one. The report will not change

gitcruz:
Thanks both for this review!!!

@diegomics  we will try to include the repeat extensions for future EAR submissions. Do you have any recommendations on how to do it? early on I just ran RepeatMasker without any specific database...

PR #216, opened 2025-09-02
Add Hypericum elodes

ldemirdj:
## Assembly review request

- ToLID: ddHypElod1
- Species: Hypericum elodes
- Project: ERGA-BGE
- Affiliation: Genoscope


jesgomez:
Yes

auryjm:
Hi @tbrown91 and @ldemirdj,

I’m pasting below an exchange we had with the sample ambassador about the Hypericum genome assembly:
_"[...] several Hypericum assemblies are already available, with assembly sizes ranging from 338 Mb to 1,541 Mb. This variation might be due to different ploidy levels. Similarly, the number of chromosomes in this genus appears to vary significantly, from 9 to 20 => https://www.ncbi.nlm.nih.gov/datasets/genome/?taxon=55962&assembly_level=2:3
Given these observations, I would expect that the assembly size is correct, potentially due to a different ploidy level similar to that observed in H. androsaemum, H. pulchrum, H. hircinum, and H. hirsutum.[...]"_

Please note that, for example, _Hypericum perforatum_ also shows a high rate of duplicated BUSCOs, which may indicate polyploidization events. Be careful when removing suggested allelic contigs or regions, as they might actually correspond to polyploidization events!

Jean-Marc

diegomics:
Hi all, I just remember that you may find interesing to try [Inspector](https://github.com/Maggi-Chen/Inspector) here since Hi-C data is not available, just a comment :)

auryjm:
Thank you @tbrown91 ,

I believe it is important to submit a 1n genome, but we cannot exclude the possibility that in the 1n genome all chromosomes are duplicated. If you then look at the 2n genome, you should observe four copies of each region.

For example, Brassica napus contains two genomes, A (B. rapa) and C (B. oleracea). The 1n genome to be submitted is the one composed of 10 A chromosomes and 9 C chromosomes and obvisouly BUSCOs duplicated score is high but this is expected.
It is sometimes not very easy to distinguish between polyploidy and alleles! I remember that for Androsace I did it based on Smudgeplot and SNP detection.

We will have a look, as another plant genome for BGE is in the same situation, except that for that one we do have Hi-C data.

PR #221, opened 2025-09-05
Add Carabus staehlini

ldemirdj:
## Assembly review request

- ToLID: icCarStae2
- Species: Carabus staehlini
- Project: ERGA-BGE
- Affiliation: Genoscope


diegomics:
Ok

ldemirdj:
Hi everyone,

Here is the updated EAR report. I’ve made the requested changes and also performed several alignments.

First, with *Carabus problematicus*, which has an X chromosome that maps well to our chromosome 1. Given the drop in coverage, we can confidently say that this is the X chromosome.

As for the Y chromosome, I did another alignment with *Carabus banonii*, which has a small chromosome, but nothing mapped to it. I also ran other tests, but none were conclusive, so I left it as is, since it will likely be difficult to identify the Y chromosome due to its small size.

Best regards,

PR #224, opened 2025-09-10
Add Zygaena laeta

andar27:
## Assembly review request

- ToLID: ilZygLaet5
- Species: Zygaena laeta
- Project: ERGA-BGE
- Affiliation: Genoscope


additive3:
OK

epaule:
Yes

epaule:
I think I would do some changes:
* SUPER_11 split it at 10.9 where the telomer dot is and invert 10.9-end, then you can trim off (original)end - gap at 13.18 as hapdup
* SUPER_13 split 11.5-end and move it to the start of the chromosome
* SUPER_19_unloc_1 would fit very nicely in reverse into the gap at SUPER_19 2.1
* similar SUPER_19_unloc_2 goes to SUPER_19 at 8.5
* then move SUPER_24 7.83-7.86 inverted to the gap at SUPER_18 8.42
* SUPER_26 split at the telomer dot at 4.47 and move it to the other end

Generally if you would have a telomer track, you probably would have seen a lot of these, but as we don't have that, the weak telomer cross-signal. are what i would use as guide.

andar27:
Hi @epaule, I made the changes that you suggested and updated the EAR report and Pretext map.

Best,
Adama

PR #229, opened 2025-09-15
update Macrothele cretica 

EmilieTeo:


EmilieTeo:
Hi @diegomics 
I have aligned the old assembly with this one, there's a very good correlation between each chromosomes. I could see a few inversions on the dotplot but couldn't necessarily see it on the map. It's also possible that the errors were made on the first version of this assembly since the signal on the map was sometimes really hard to interpret. 

Here's the dotplot of the alignment
image

There are indeed 3 scaffolds that are still shown as bacteria on the tapas plot, they are not the ones mentioned in the curator's notes. Only one gene was found on those scaffolds so they were not removed in our pipeline. 
I had also noticed the differences in the BUSCO values, maybe you're right and the differences can be explained by the other version, but it was a bit surprising because this map was much easier to curate compared to the first version so I wasn't expecting lower values

Best, 
Emilie

PR #237, opened 2025-09-24
idFanCani4 EAR upload

epaule:
## Assembly review request

- ToLID: idFanCani4
- Species: Fannia canicularis
- Project: ERGA-BGE
- Affiliation: Sanger


talioto:
Yes

talioto:
First impressions. Blobtools shows one large sequence with higher coverage. From the pretext I see that this is SUPER_6, which contains many repetetive sequences. (repeat track and coverage track confirm this), half of which contact non-specifically with centromeric sequence from other supers. the other half of the "SUPER_6" seem to have non-specific contacts with other supers and stronger contacts with the "centromeric" seqs within SUPER_6. This will require some attention, but SUPER_6 does not appear to be a real chromosome, but more of an agglomeration of repeats.
Screenshot 2025-09-26 at 3 02 28 PM

epaule:
in the original merged map, there are two copies for the SUPER_6 
Screenshot 2025-09-29 at 13 17 04 . The exact order of the repetitive bits on the right arms is debatable, but i think the left arm looks fine and conserved between the two haplotypes. The diptera BUSCOs are also only on the left arm (between 1.6 and 39 Mbp). If you want I can unloc the repeat scaffolds, to better express the uncertainty.

talioto:
HAP1_scaffold_19 could be unloced to SUPER_6. Maybe some others in the chaff but I wouldn't worry about them. The pretext map is noisier this time and the coverage track is messed up, but other than that, I'd say you're good to go.

talioto:
Even with mq0, I can't make out the order of them well. Scaffolds should be ordered and oriented, otherwise place them unlocalized. If you want to leave them in the super scaffold as they are, then you need to explain that the order is unsure.

epaule:
They are just 11 contigs, ordered as well as i could. So i would opt to put in a disclaimer about the order and orientation into the genome note... also then we don't have to rebuild the EAR.

PR #242, opened 2025-10-01
initial commit Laparocerus anagae

MartinPippel:
## Assembly review request

- ToLID: icLapAnag1
- Species: Laparocerus anagae
- Project: ERGA-BGE
- Affiliation: SciLifeLab


gitcruz:
Dear @MartinPippel 

Congratulations, this is an excellent diploid assembly. However before approval I'd like to discuss the following points:

1. I think that a clear haplotig remains trapped into a contig in hap2, at superscaffold h2su10 (start at 4Mb position). I think it could be the product of a miss-assembly and took it off and put at the end in snap mode. Personally I'd cut it out and tag it as haplotig. [Please find my savestate ](https://ccnag-my.sharepoint.com/:u:/g/personal/tyler_alioto_cnag_eu/EXcefi5JD01Cg85GMHH26HsBt3pibKSY9gvnEj1-JbZHng?e=4HYoNM)here and let me know what you think?

2. I totally agree with h1su11 being the sex chromosome. WRT taggin' it I suggest tagging it as X (you already wrote  sex is X0 in the report). I found this [publication supporting X0 ](https://pubmed.ncbi.nlm.nih.gov/39585913/) systems for other beetle genomes. In addition you could also run a WGA against the [chrom-level assembly ](https://www.ncbi.nlm.nih.gov/datasets/genome/GCA_965642125.1/) of a distant species Catops nigricans (divergence 222MYA), that already contains an X chromosome. So that naming seems appropriate to me.

3. Could you tell me what you think is happening at the beginning of super9. There is something going on in both haps. But in h1su9 I see signals connecting h1su8 (¿haplotigs or subtelo repeats?) and also with h1su11 (it seem to coincide with real duplication). See the screenshot below:

hap1_icLapAnag1_h1su9_arm1


Let @diegomics have a look too.

Best regards,
Fernando

MartinPippel:
Thank you both @gitcruz, @diegomics 
We don't have bionano.
How do we continue now? Do you trust me that I rename the sex chromosome before ENA submission or do I need to rerun the whole pretext curation pipeline? Though, I would prefer the former ;)

MartinPippel:
Thx @diegomics. I will continue with the ENA upload!

diegomics:
Yes, if the change is not going to show in the EAR, just proceed. We trust you guys ;P

PR #243, opened 2025-10-01
inital commit Cladocora caespitosa

MartinPippel:
## Assembly review request

- ToLID: jaClaCaes1
- Species: Cladocora caespitosa
- Project: ERGA-BGE
- Affiliation: SciLifeLab


additive3:
OK

additive3:
Hi @MartinPippel 

Nice assembly and good to see that the Hi-C worked well, not always the case with cnidaria!
My only query are virtually all the shrapnel contigs, these look suspicously like contamination. Would you be able to run some further checks.

PR #247, opened 2025-10-03
iyAntRoge3 EAR

gbdias:
## Assembly review request

- ToLID: iyAntRoge3
- Species: Anthophora rogenhoferi
- Project: ERGA-BGE
- Affiliation: SciLifeLab


diegomics:
ok

bistace:
Yes

bistace:
To get the number of interventions per Gb, we use the `curated.log` file that `pretext-to-asm` generates as you mentioned. We look at this line:
```
Curation made 0 cuts in contigs, 11 breaks at gaps and 10 joins
```
Then we sum the cuts, breaks and joins and divide by 1,000,000,000. I think that this does not include unlocs because they do not introduce modifications to the sequence of the assembly.

I will take a look at the updated map and EAR.

diegomics:
One thing that was confusing from the PDF was "The assembly was produced with hifiasm in Hi-C mode", but since the hi-c data is from a different individual, my understanding is that this was not the case, right? Please can you clarify @gbdias

diegomics:
Ok let's merge it, cheers on another one!

PR #266, opened 2025-10-22
Support MerquryFK output in EAR

mahesh-panchal:
Hi, I added support for MerquryFK output in the EAR report generator.
Is this of interest?


diegomics:
After a long time, this is finally landing, thanks Mahesh. I'm closing this since the imminent EAR update is capable of using merqury and merquryFK (also compleasm in addition to busco).

PR #350, opened 2026-02-11
Add  Pseudanchomenus aptinoides

SMangenot:
## Assembly review request

- ToLID: icPseApti1
- Species:  Pseudanchomenus aptinoides
- Project: ERGA-BGE
- Affiliation: Genoscope


tbrown91:
ok

tbrown91:
ok

JoannaCollins:
Hi @ldemirdj 

I took a couple of days leave so only just getting to it now.

JoannaCollins:
Hi @SMangenot,

Thank you for sending the updated map :-), and sorry for the delay in getting back to you.

I asked Jo Wood (@additive3 ) to take a look at this one as i was still concerned about chromosome numbers and the data is tricky to interpret. Having assessed this yesterday he suggested that a number of the units currently tagged as chromosomes are likely not chromosomes (chrs 3, 5,10, 17, 21).  For 10, 17 and 21 you can see that they all hit heavily to the repetitive scaffolds in the shrapnel this repeat should be unasigned. For chromosomes 3 and 5, these are half cov and Jo suspects these repeats may be Y linked.

In order to try and fully resolve the assembly would it be possible for you to send the precurated combined pri/alt map (both normal mq0 and mq10). If it doesn't already exist it would be really good if you are able to generate this to try and unpick which scaffolds are part of which chromosomes.

Thanks very much

Jo

auryjm:
Hi @JoannaCollins , @additive3, @SMangenot and @tbrown91 ,

For this tricky assembly, unfortunately we do not have the primary/alternate assemblies, as the scaffolding was performed from the collapsed assembly. I had a look at the map and, apart from chromosomes 3, 5, 10, 17, and 21, which I agree appear to correspond to repetitive regions, the others look good and of sufficient quality.

I propose moving these chromosomes to the shrapnel as unplaced sequences, adding some explanations in the assembly description, and finalizing the assembly so that we can move forward. What do you think?

Jean-Marc

JoannaCollins:
Hi @auryjm, @SMangenot and @tbrown91 ,

@additive3 and I have had a discussion about this assembly today and both agree that given the quality of the data your proposal to unassign current chrs 3, 5, 10, 17, and 21 and add a comment to the assembly description is the most logical thing to do in this scenario.

Best Wishes

Jo

JoannaCollins:
Hi  @SMangenot,@auryjm and @tbrown91,

Thanks for uploading the updated map :-).

I have had a look through and there are just a couple of things to correct.

It looks like there my have been a painting/tagging error as in the assembly there is a SUPER_16 and 3 SUPER_16 unlocs but all of these are tiny scaffolds so these need unpainting/tagging. The observed chr number in the assembly report will then also need updating.

I would unassign SUPER_10_unloc_1 as i don't think we can say it defintely belongs with SUPER_10.

The first 2 chromosomes are half coverage, you might want to consider assigning the sex chromosomes if you can find a close releative to align the data to.

Any questions please let me know

Best Wishes

Jo

tbrown91:
@SMangenot The only thing I would say about this one is in relation to @JoannaCollins' about the first two scaffolds. These seem to be very clearly at lower coverage than the rest of the assembly, so I would lean towards tagging them as sex chromosomes.

Other than that, I think we can probably wrap this up. @additive3 please have a look through in case you want to add anything

JoannaCollins:
Hi @SMangenot,@auryjm and @tbrown91,

Thanks very much for sending on the updated map :-).

Having had a chat with @additive3 we feel its ok not to assign the reamaining half coverage chromosomes if we can't be certain of their assignment.

There are 2 remaining issues to fix and then this can be signed off.  

SUPER_10: 22.31-24.47Mb - I think this was previously SUPER_10_unloc_1 which i mentioned should be unassigned (and left in the shrapnel as we can't be sure where it goes) has now been jonied to the end of SUPER_10. This needs removing from the end of this chromosomeas you can see there is telo signal where the join has been made.

SUPER_12: needs breaking @ the gap 7.08Mb as there is telo signal here indicating that SUPER_12 splits into 2 chromosomes. In a previous iteration of the map these were 2 seperate chromosomes so at some point have been joined but the telo evidence does point to them being seperate entities.

I have attached a savestate indicating with waymarkers the regions that should be broken off.

Thanks very much

Jo
[icPseApti1_hr.pretext.map.savestate_1_jcc_6.8.26.txt](https://github.com/user-attachments/files/30796126/icPseApti1_hr.pretext.map.savestate_1_jcc_6.8.26.txt)

JoannaCollins:
Hi @SMangenot, @auryjm and @tbrown91,

Thanks for uploading the new map @SMangenot :-).  I have had a look through and the issues from the previous itteration are now fixed which is great. However there seems to have been some unexpected additional changes:

What was originally Scaffold_23 has now been assigned as chromosome 17 (icPseApti1_17), this should be unassigned as a chromosome and placed back in the shrapnel and the final chromosome count for the assembly should be 17 rather than 18. 

The chromosome that was assigned as X previously has now been unassigned and is now called icPseApti1_6, as you are confident with the assignment of this sex chrmosome it would be good to rename this scaffold X as it was before.

Any questions please ask,

Best Wishes

Jo

PR #352, opened 2026-02-16
add qqNelTrog

jesgomez:
## Assembly review request

- ToLID: qqNelTrog
- Species: Nelima troglodytes
- Project: ERGA-BGE
- Affiliation: CNAG


jesgomez:
Hi @MartinPippel @tbrown91 ,

I had decontaminated with FCS only because I was having trouble with blobtoolkit but I finally managed to get the results and I detected some more contaminants. Thanks to that, I managed to get rid of the first weird peak. Now we are finishing the manual curation and we'll update the pdf soon.

gbdias:
Hi @jesgomez ,

Martin is a bit busy with another project and asked me to take a look at this version of the assembly.

Your stats look excellent and the chromosome assignments look correct to me. There's a few off-diagonal signals but I couldn't resolve them with edits, so perhaps these represent differences between individuals used for sequencing?

There is one large unassigned scaffold that I think could go to Super8 (details below, let me know what you think). In any case I would be comfortable approving the assembly as it is.

H1_Scaf25 could be unloc of H1_Super8, or possibly in the gap at 26.5Mb.
similarly
H2_Scaf18 could be unloc of H2_Super8, or go in the gap at 27.4Mb.

jesgomez:
Thanks @gbdias. I agree about placing the two scaffolds mentioned above in the suggested gaps, thank you. Let's see if @tbrown91 has any additional suggestion and if not, I'll produce the final files.

tbrown91:
I'm very happy. Thank you both

MartinPippel (approved):
Thank you @gbdias and congrats to @jesgomez!

PR #368, opened 2026-02-23
icParSeri upload

msozzoni:
## Assembly review request

- ToLID: icParSeri5
- Species: Parapropus sericeus
- Project: ERGA-BGE
- Affiliation: UNIFI


tbrown91:
ok

CaroB-M:
Yes

talioto:
@tbrown91 , please finish or @diegomics , assign a new reviewer.

auryjm:
Hi @msozzoni ,

Thank you for this new assembly. I have taken over the review, as Caroline has been out of the office for the past few months.

This genome is highly fragmented, with a low contig N50, which makes it difficult to validate the internal chromosomal organization. I think these limitations  should be documented in the assembly description accompanying the submission so that users are aware of the assembly context and its remaining uncertainties. I performed a number of inversions and several scaffold relocations that, overall, substantially improved the Hi-C contact map. 

My main concern is the chromosome number. You painted 17 autosomes plus the X chromosome, but I believe that several chromosomes should be fused. After reviewing the Hi-C data, I ended up with 12 autosomes plus the X chromosome, which is much more consistent with other Leiodidae species. For example, Graciliella metohijensis has 12 autosomes plus the sex chromosomes.

Please find my saved state attached. I think it would be worthwhile to go through the Hi-C map again after applying these proposed changes, and also to compare the resulting assembly against a closely related reference genome.
[icParSer5.Hap1.1.pretext.savestate.txt](https://github.com/user-attachments/files/30260844/icParSer5.Hap1.1.pretext.savestate.txt)

One final point: several unloc scaffolds have been moved to different chromosomes. I think they should be renamed before generating the final assembly; otherwise, they will retain the names of their original chromosomes.

msozzoni:
Hi @auryjm 

Thank you for the review. I agree with all the changes you have made and will go through the HiC map again after I apply your changes. I will also add a comment about the low N50 on the final submission as suggested. 

I too was concerned about the number of chromosomes, but I was wary about checking with an alignment to a close species, as it was discouraged in another Leiodidae species I have assembled (Anthroherpon ganglbaueri). If you think an alignment or at least a comparison will be helpful, I will perform it and get back to you with the results.

PR #376, opened 2026-02-26
Ulmus laevis EAR

diegomics:
## Assembly review request

- ToLID: drUlmLaev1
- Species: Ulmus laevis
- Project: ERGA-BGE
- Affiliation: IZW


diegomics:
Sorry for the delay guys, I'll try to work on this on the coming days 🙏

additive3:
Hi @diegomics @JoannaCollins 

Any progress with this assembly?

diegomics:
I have a student working on it but it's delayed. I may handle it mayself

additive3:
Np.. and from my PoV no rush.

diegomics:
Hi @JoannaCollins @additive3
Please find [here](https://box.fu-berlin.de/s/gs2j2rc9yAXBGM2) the hap1hap2 contact map and the savestate.
The curation is being handled by @henryspaul (he's happily learning a lot)

JoannaCollins:
Hi @diegomics and @henryspaul,

Apologies for the delay in getting back to you with this. Thanks very much for sending the merged curated map with the corrections. I have had a another look through and made some edits as well. With regard to the regions that we suspected might be hap dups and belonged in the other hap we can see in the merged map that actually these are just duplicated regions from the same hap. Looking at the repeat tracks we can see that these duplications are all in regions of low complexity repeat where we do typically see a drop in coverage compared to the euchromatic sequence. I can also see that over some of these regions coverage is stacked indicating that the region is collapsed. Taking these aspects into account and as there is some ambiguity i think in this case it makes sense to leave in the duplicated regions. I have made a savestate where i have painted up the chrs. One thing to note from looking at the savestate you sent is that if you tag something as haplotig/falseDuplication you do need to move it to the end of the scaffold otherwise you will run into issues when you try to remake the curated.fa.

Any questions please don't hesitate to ask

Best Wishes Jo
[drUlmLaev1.hap1hap2.cur.20260223_tracks.pretext.savestate_1_jcc.txt](https://github.com/user-attachments/files/29040372/drUlmLaev1.hap1hap2.cur.20260223_tracks.pretext.savestate_1_jcc.txt)

PR #381, opened 2026-03-01
Add qqGyaAnnu4 EAR

tbrown91:
## Assembly review request

- ToLID: qqGyaAnnu4
- Species: Gyas annulatus
- Project: ERGA-BGE
- Affiliation: IZW


tbrown91:
Hi both,

Here is the link for the maps: https://box.fu-berlin.de/s/bTXpFaCopaDbJNs

Apologies for the delay

talioto:
Just to let you know that I'm now actively taking a look at this one. I will make some comments here and you can look at my [savestate](https://ccnag-my.sharepoint.com/:u:/g/personal/tyler_alioto_cnag_eu/IQC0OvyU0mvcSpBN6Ms5dNXPARnHffsRCREDUnseSHdGdpw?e=Igltah).

- SUPER_6 and SUPER_12 look like they can be joined. BUT, this would introduce a telomere in the middle. If we care about telomeric repeats in the center of supers, then we should probably revise SUPER_1, SUPER_2, SUPER_3 and SUPER_10 so that they don't join on telomeres (I've broken them up in my savestate). Can you confirm the telomeric repeat in the sequence and make sure it is repeated in tandem and not just enriched? TIDK uses a frequency in window approach, but on closer inspection, sometimes we can rule out some of the peaks.
- The repetitive contigs in SUPER_4 should probably be unloc. I don't see the evidence for their order and orientation. Maybe if you shared a map with a higher mq threshold it would help? Plenty of the shrapnel would appear to go here, too, but I'm worried that this is just an "attraction" point -- it could actually be placed in other chromosomes, too. I would break SUPER_4.
- weak support at SUPER_5:28.66. Not sure of orientation here. I changed it up but low confidence.

additive3:
@tbrown91 @talioto 

Nice to see a Harvestman genome!

Agree that SUPER_2, SUPER_3, SUPER_4 and SUPER_10 should be split.
I'd guess the repeat on SUPER_4 is terminal rDNA
SUPER_6/SUPER_12 think this is just a reflection of similar repeats giving elevated Hi-C signal and these should be separate.

SUPER_1.... hmmmmmmmm it is also possibly 2 joined because of all the repeat need to look more indepth at it.

Out of interest what was the identified telomere motif?

talioto:
Seems like @additive3 and I are more or less in agreement. The ball's in your court, @tbrown91

tbrown91:
Hey. Can you have a look at this?

[qqGyaAnnu.tracks.pretext.savestate_1.gz](https://github.com/user-attachments/files/27209988/qqGyaAnnu.tracks.pretext.savestate_1.gz)

Telomere motif used was TTAGG

I'll be AFK again for a week or so now. I'll do what I can when I'm back

talioto:
What is the MQ cutoff for this map? I feel like there is a lot of repeat attraction and it's hard to tell what really is contacting what. Also, maybe you could supply the assembly fasta file. I'd like to look a little closer at some sequence features.

talioto:
@tbrown91 , can you please just finish this genome. If Jo approves, I'm fine.

tbrown91:
@additive3 Chr1 is a mess, but I think the rest is looking in good shape.
If you can do anything, let me know. Otherwise hopefully we can Wrap it up

PR #392, opened 2026-04-15
add qmTriMatu1_EAR.pdf

zilov:
## Assembly review request

- ToLID: qmTriMatu1
- Species: Trichoniscus matulici
- Project: ERGA-BGE
- Affiliation: Sanger


diegomics:
Ok

gbdias:
Yes

gbdias:
Hi @zilov ,

Thanks for the assembly. The stats look fine and I don't have any suggestion regarding the curation. 

Just one question, what was the N50 on the HiFi reads before assembly? The contig assembly seems a bit too fragmented. But I haven't done any isopod assembly so maybe there's something with the genome structure as well that could influence the contig N50.

zilov:
Hi @gbdias, thanks for the feedback. I checked the HiFi read N50, and the average across our three samples is around 13 kb. I think the fragmentation you're seeing is expected given the genome size, and as you suggested, there could be specific structural features at play here.

gbdias (approved):

PR #395, opened 2026-05-19
add hrDivMult2 EAR

zilov:
## Assembly review request

- ToLID: hrDivMult2
- Species: Diversibipalium multilineatum
- Project: ERGA-BGE
- Affiliation: Sanger


diegomics:
ok

gitcruz:
Yes

gitcruz:
Yes

gitcruz:
Dear Danil @zilov, 

Congratulations this is a very good assembly for a Platyhelminth! You obviously went through extensive curation with it. 

However, this is what I saw:

1. 4 small gaps at SUPER_1@134,596K, SUPER_2@136,247K, SUPER_3@20,356K and SUPER_4@98,210K are part of the centromeric region containing a gap in SUPER_5. They show stronger contacts there and the diagonal signal of the original scaffolds improves after moving them. The gap from SUPER_4 is hard to localize (I left it as unloc). Other option would be left the four sequences unplaced into the "chaff" bin.
2. In my opinion, SUPER_1 the contig around 143.5M-144.56M I think it flips. 
3. I localized or placed some of the small scaffolds in the "chaff".

In the link below you can download a save_state and go over the map, please let me know what you think.

[hrDivMult2_normal.pretext.savestate_fc_1](https://ccnag-my.sharepoint.com/:u:/g/personal/tyler_alioto_cnag_eu/IQAyrP9bFsOeRIQLK_1aKDrrARtby-NfQ3rquzF22NXIX5M?e=d1YJ92)

@diegomics do you have any other comments?

Hope it helps,
Fernando

diegomics:
Thanks Nando, I don't have further comments than yours regarding the contact map. Please let us know what you think @zilov.
Just for the sake of tracking, please confirm the mollusc hit as false positive (it's not stated in the contamination notes)

gitcruz:
True, based on its length corresponds with SUPER_3 and I assumed is a false hit.

> Thanks Nando, I don't have further comments than yours regarding the contact map. Please let us know what you think @zilov. Just for the sake of tracking, please confirm the mollusc hit as false positive (it's not stated in the contamination notes)

zilov:
Hi @gitcruz,

Thank you so much for the detailed feedback and for preparing the save_state, that's genuinely helpful.

I've reviewed your comments carefully:

1. Centromeric gaps (SUPER_1/2/3/4) — I'd prefer to leave these in their current positions. The stronger contacts you observed at those loci are, in my view, more likely to reflect contig-end artefacts than true misplacement:

   - Hi-C signal is usually noisier at scaffold junctions, so decreased contacts near contig boundaries are expected and don't necessarily indicate a curation error.
   - These regions correspond to repetitive sequence, which typically shows shared signal with many other repeat-containing loci — making placement calls genuinely ambiguous.
   - On SUPER_5, the centromeric repeat itself shows limited inter-chromosomal signal, and adding these fragments does not meaningfully strengthen contacts to the rest of the chromosome.
   - When I return the fragments to their original chromosomes one by one, I see a long tail of contacts back to other parts of the original scaffold — which is more consistent with a contig-end artefact than a misplaced contig. (see screenshot 1)

Given the Hi-C bin size, a (1-2-3)-pixel positional difference also falls well within the noise threshold. I think the current placement is defensible, and moving these fragments introduces more uncertainty than it resolves.

2. Flip at SUPER_1 ~143.5–144.56M — I've looked at this carefully and I don't think the flip is supported:

   - The out-of-diagonal contact tail for this region runs in a direction consistent with the current orientation, not the flipped one.
   - The region at SUPER_1 ~9.88–10.24 Mbp, which lies close to the diagonal, also shows elevated signal with this contig — and that signal pattern is more consistent with the current placement than with a flip. (screenshot 2 - original state)

3. Chaff scaffold placements — I agree with your suggestions. 

One thing worth flagging: our server is currently in a service, so the update of the map could take around a week or two. As alternative to map update I can suggest manual unlocks renaming using the table below:

HAP1_SCAFFOLD_42, HAP1_SCAFFOLD_161 - SUPER_1_unloc_1, SUPER_1_unloc_2
HAP1_SCAFFOLD_35, HAP1_SCAFFOLD_154 - SUPER_2_unloc_1, SUPER_2_unloc_2
HAP1_SCAFFOLD_113 - SUPER_3_unloc_1
HAP1_SCAFFOLD_29, HAP1_SCAFFOLD_40 - SUPER_4_unloc_1, SUPER_4_unloc_2
HAP1_SCAFFOLD_76, HAP1_SCAFFOLD_134  - SUPER_5_unloc_1, SUPER_5_unloc_2

If that works for you, we could proceed with submission straight after the renaming. I'll update the chromsome list with the unlocks as well.

Let me know what you think.

Best,
Danil

screenshot_1_super_4_repeat
screenshot_2_flip_piece_original_state

gitcruz:
Hi @zilov,

I went back to the flipped contig inside SUPER_1 (143-144.8Mb) and is true that the rest of the off-diagonal signal looks much better in the original orientation (leave it like that), there's a strong signal due to repeat attraction that i put together with the flip but the rest gets worse.

I reviewed these 4 sequences of ~200-300K (that contain some kind of centro repeat) and I agree with you. I was a bit conservative and inclined to leave them unplaced. However, when I compared them all against SUPER_5 (screenshot_1) and their respective scaffold of origin, the signal is larger and more continuos there (example SUPER_2 screenshot_2). You did a good job.

I don't think is necessary to regenerate the map but can produce a new fasta with the placed unlocs in it. Please note that I also localized a few sequences two, so perhaps is better to take my savestate and put the 4 sequences in their original spot.

screenshot1_all_vs_SUPER_5

screenshot2_all_vs_SUPER_2

I'll approve so you can upload the assembly to ENA if it's ok with @diegomics.

Thanks,
Fernando

gitcruz (approved):

diegomics:
I agree, please just update the EAR with the new fasta metrics, no need to regenerate the contact map. I'll merge this PR after.

zilov:
Hi @diegomics,

Since we're not making any changes to the sequences themselves — only renaming headers of the shrapnel/unloc scaffolds — the assembly metrics in the EAR (contig/scaffold N50, total length, etc.) will remain unchanged. I don't see a reason to update the file, unless you had a specific metric in mind that I'm missing?

Happy to update if needed, just let me know.

Best, 
Danil

gitcruz:
Hi @zilov 

As far as I remember, I placed 1-3 sequences next to a gap, that changed the total number of scaffolds, right?. Anyway, I agree that this will not affect to the general assembly metrics. Just keep it in mind while writing the genome note.

@diegomics ?

Best regards,
Fernando

zilov:
Hi @gitcruz, 

I see now, I thought that we agreed on renaming of all of the shrapnel pieces as unlocs + genome note mention without adding them within chrs. I'll add them to the chrs and update metrics than.

Thanks, 
Danil

gitcruz:
I don't know if you used my savestate. Here (https://github.com/ERGA-consortium/EARs/pull/395#issuecomment-4591041325), I mentioned "Please note that I also localized a few sequences two, so perhaps is better to take my savestate and put the 4 sequences in their original spot." 

Honestly, I don't remember how many and where I placed them right now...

Best,
Fernando

zilov:
No worries, I placed them according to the savestate, I'll update the EAR shortly

gitcruz:
Ok, so then you see a small difference in the number of total scaffolds?

Thnx

zilov:
Hi @diegomics, @gitcruz,

I've updated the stats of the assembly after unlocs placement. Please take a look :)

Thanks,
Danil

gitcruz:
Hi @zilov 

After looking at review stats, the assembly span increased due to to the introduction of 47.6 gaps? and the number of scaffolds got dramatically reduced...I don't remember that we have placed that many sequences and doing that many joins...

Before review -> Reviewed
TotalLen:845,560,297 -> TotalLen:846,494,783 (increased 934,486 bp)
Scaffolds:236 -> 190
TotalGapLen: 39,500-> 49,000 (9,500bp: 47.5 new gaps of 200bp???)
Contigs: 457 -> 456 

It seems that we missed 1 contig and the amount of sequence gained is higher than the total number of gaps. How that can be?

Did you ran gfatstats on the reviewed assembly or simply edited the numbers?

Did pretext-to-asm reported any errors/warnings? 

why we have lost a contig?

Sorry, I don't unerstand it.

Thanks
Fernando

zilov:
Hi @gitcruz,

1. I've checked your comments on unlocs, than found them in original map (its both hap1 and hap2 map).
2. I've clustered the remaining shrapnel to see if some of them are uniquely matching to chromosomes (since we do edits to fasta than some others could be added too).
3. Than found that lots of repeats like hap1_scaffold_100 (see screenshot 1) mapping uniquely to chr1.  For some other repeats I've added more seqs that showed clear signal of join so that increased total number of gaps in the assembly.
4. Another thing that I found is a large block in hap2 scaffolds (HAP2_SCAFFOLD_8 0.00 Mbp - 1.16 Mbp) that was not present on hap1 chromosome 2 but had a copy on hap2. So I've added it to hap1 and it increased the total length of the hap1 assembly. (screenshot 2).
5. Than I've checked for other scaffolds that might be assigned to hap2 by hifiasm but not represented on hap1. No large scaffolds were found.
6. Than I've took a closer look on HAP1_SCAFFOLD_29 and HAP1_SCAFFOLD_176, on original map it seems to be a third copy of the region already presented on hap1 and hap2, so I decided to tag it as as false-dup and rm from the assembly. That decreased number of contigs.
7. After all edits I've run pretext-to-asm, gfastats and busco and updated the table in pdf.
8. Now I'm running hic remapping, I can update the link in EAR once it's ready so you can take a look

image

image

Note that we work from the original dual-haplotype map, not the remapped one, so edits on your side would need to be independently verified and re-applied on mine. We do this to keep track of all edits of original assembly. Would it make more sense to share the original save state in addition to the remapped map? Do you have access to original maps?

Let me know what you think.

gitcruz:
Hi @zilov 

Let's roll this back, you have submitted a primary assembly for EAR review. If I got it right you basically did a diploid curation, and the representativ haplotype (hap1+ some missing stuff from hap2) is what you kept and will upload as "primary" (thus discarding hap2 leftovers). That's ok with me. However, what I can't folow up is this going back and forward to a diploid map that you haven't shared with us (@diegomics  as supervisor and myself as reviewer). 

Also the fact that you have imported extra sequence from hap2 explains the sequence gain. Honestly, you should really understand that this is very confusing and you should at least explained in the notes.

Please share your final diploid map, and explain what are you keeping as primary so we can review this once again. To me is ok with the map plus your latest savestate. 

Thanks
Fernando

zilov:
Hi @gitcruz,

Apologies for the confusion. We always curate from the diploid map internally — this is part of our standard process, not something specific to this assembly. While working through your savestate I found additional sequences worth placing and two false duplicates, so I included those changes as well. I should have flagged this before updating the EAR.

[Link to original map and savestate](https://drive.google.com/drive/folders/1so7jY8_38lA39XZnNK8kNr_YIlPYKTll?usp=sharing), happy to walk through any specific edits if something looks off.

Thanks,
Danil

diegomics:
Thanks for the clarification @zilov, now I get it :)
Like @gitcruz, I was also not following.
Merging now, thanks again to both

gitcruz:
Ok, @diegomics merged the PR. 

Anyway @zilov, next time please share the maps you used for curation and make it clear.  Congratulations is a good assembly of a complex genome.

Have a nice weekend

zilov:
@gitcruz, @diegomics,

Thanks for the patience and for merging — and sorry again for the confusion. I assumed that we are on the same page.

If it's still relevant and of interest, I'd be happy to prepare a short write-up or talk on our workflow so we can sync up the two sides and avoid this kind of back-and-forth in future submissions. Happy to update our guidelines too if that helps things move faster through QC.

Have a great weekend, and thanks again for the thorough review — submitting the assembly now.

Danil

gitcruz:
Thanks, no worries @zilov the assembly looks good. Just go ahead with the upload now that the PR is merged.

PR #405, opened 2026-08-11
Fix comment handling, review detection and reviewer bookkeeping

arash77:
Builds on #404, so that change shows up here too. Review #404 first, then this one; the second commit is the only new part.

**A drive-by comment review no longer blocks a PR.** `get_reviews()` also returns COMMENTED reviews, which any user can leave on a public repo and which cannot be deleted afterwards. Both `find_reviewer` and `comment` treated that as "already reviewed", so one stray comment review stopped a PR from ever getting a reviewer, with no error and no label. New `_has_binding_review` helper counts only APPROVED and CHANGES_REQUESTED.

**CLEAR now stops after clearing.** It had no exit, so control fell through into the supervisor confirmation branch, which re-added the ERROR! label CLEAR had just removed, posted "Invalid confirmation!" and exited 1.

**OK is parsed like Yes and No.** It used a bare substring test over the whole comment, so "Looks good to me", "took a look" and "broken" all counted as OK, and so did quoting the bot's own message back at it. Both paths now share `_first_reply_line` and use word boundaries.

**`approve_reviewer` checks the review that fired the event** instead of `get_reviews()[0]`, which is the oldest review on the PR. Any earlier review by somebody else suppressed the thank-you comment, and `closed_pr` relies on that comment to identify the reviewer. It also no longer crashes when no supervisor is assigned.

**`closed_pr` fixes:** no crash on a PR closed before a supervisor was assigned, the EAR PDF is resolved before either CSV is written so a missing PDF cannot leave the data half updated, and `other_participants` is keyed by GitHub ID so the roster Full Name lookup matches. It was comparing a lower-cased ID against a set of display names, so it almost never fired and the column ended up mixing real names, display names and bare logins.

**The timeout penalty applies again.** It was an `elif` on a branch that always won first, so the "Calling Score + 1" in the README never ran and a reviewer could ignore every request with no effect on their ranking.

**`select_best_reviewer` returns an empty list** instead of raising IndexError when nobody is eligible, which had made the "No eligible candidates found." message unreachable.

**Supervisor lookup is case-insensitive.** Roster IDs were compared raw against a lower-cased author, so a supervisor whose login has any uppercase could not confirm a PR. Five roster rows already have uppercase logins.

**Slack posts escape `&`, `<` and `>`** in names taken from the PR body, so a species name cannot contain `<!channel>`.

Each of these was checked against the current code first to confirm it reproduces, then against the fix. 12 checks, all failing before and passing after. The `ruff` E402 warnings in `get_EAR_reviewer.py` are pre-existing and untouched.

### DIFF
--- a/.github/workflows/1_ear_bot_pr.yml
+++ b/.github/workflows/1_ear_bot_pr.yml
@@ -10,6 +10,8 @@ concurrency:
   group: ear-bot-pr-${{ github.event.pull_request.number }}
   cancel-in-progress: false
 
+permissions: {}
+
 jobs:
   add-label-and-find-supervisor:
     runs-on: ubuntu-latest
--- a/.github/workflows/2+4_ear_bot_comment.yml
+++ b/.github/workflows/2+4_ear_bot_comment.yml
@@ -8,6 +8,8 @@ concurrency:
   group: ear-bot-pr-${{ github.event.issue.number }}
   cancel-in-progress: false
 
+permissions: {}
+
 jobs:
   new-comment:
     if: github.actor != 'erga-ear-bot[bot]' && github.event.issue.pull_request && (contains(github.event.issue.labels.*.name, 'ERGA-BGE') || contains(github.event.issue.labels.*.name, 'ERGA-Pilot') || contains(github.event.issue.labels.*.name, 'ERGA-Community'))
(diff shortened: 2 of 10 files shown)


arash77:
Superseded by #407, which contains this commit plus the reworked storage layer and the fixes for the regressions a later review found in it. Consolidating into a single PR so no broken intermediate state can land on main.

PR #406, opened 2026-08-11
Make roster CSV writes conflict-safe and quote-correct

arash77:
Builds on #404 and #405, so their commits show up here too. Review order is #404, then #405, then this one. The last commit is the only new part.

**Concurrent writes are rejected instead of clobbering.** `commit()` now takes the blob SHA the content was based on and passes it to `update_file`. It used to re-read the SHA at write time, which made every write succeed and silently discard whatever another run had committed in between. The scheduled run uses concurrency group `ear-bot-scheduled` while the comment and merge runs use per-PR groups, so they overlap freely. Commit `6434678` looks like this happening: a completed review and the whole Calling Score column were rolled back.

**`commit()` raises instead of catching and printing.** A lost update used to look exactly like a success in the job log, so none of this was visible in the Actions UI.

**Roster read and write both go through the `csv` module.** Reading used `line.split(',')` and writing used `",".join(row.values())`, so one comma in any field shifted every following column, and the header was rebuilt from dict key order rather than from the file. `add_pr` builds its row with `csv.writer` for the same reason, since species, names and institutions all come from the PR body. Checked that the current `reviewers_list.csv` round-trips byte for byte, so this does not rewrite the file.

**`update_reviewers_list` raises on an unknown ID.** It used to skip silently, print `Updated the reviewers list for <id>` and commit an unchanged file.

Six checks, failing before and passing after, including a simulated concurrent write that the old code applied and the new code rejects.

Not included: the `Working PRs` column is currently out of step with reality. It totals 13 against 8 real in-flight assignments. This PR stops the drift getting worse but does not correct the existing numbers, since that is data about real reviewers. Current versus actual, from the open PRs: auryjm 1/2, jesgomez 0/1, gitcruz 2/1, talioto 2/1, and bistace, CaroB-M, SarahPelan, tommathers and n-equals-one each hold 1 while being the requested reviewer on no open PR. Happy to push that correction if you want it.

### DIFF
--- a/.github/workflows/1_ear_bot_pr.yml
+++ b/.github/workflows/1_ear_bot_pr.yml
@@ -10,6 +10,8 @@ concurrency:
   group: ear-bot-pr-${{ github.event.pull_request.number }}
   cancel-in-progress: false
 
+permissions: {}
+
 jobs:
   add-label-and-find-supervisor:
     runs-on: ubuntu-latest
--- a/.github/workflows/2+4_ear_bot_comment.yml
+++ b/.github/workflows/2+4_ear_bot_comment.yml
@@ -8,6 +8,8 @@ concurrency:
   group: ear-bot-pr-${{ github.event.issue.number }}
   cancel-in-progress: false
 
+permissions: {}
+
 jobs:
   new-comment:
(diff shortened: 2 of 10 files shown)


arash77:
Superseded by #407. The conflict-safe write from this PR was itself found to cascade ERROR! labels on the failure path; #407 reworks it into a retry and folds in this change.

